PR #1, opened 2015-12-20
Add message decorators

sagikazarmark:


sagikazarmark:
see php-http/httplug#98

PR #2, opened 2015-12-20
Add message factories

sagikazarmark:


sagikazarmark:
see php-http/httplug#98

PR #3, opened 2015-12-20
Add authentication

sagikazarmark:
see php-http/httplug#98

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,11 @@
             "Http\\Message\\": "src/"
         }
     },
+    "autoload-dev": {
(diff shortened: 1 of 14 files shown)


spec/AuthenticationBehavior.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace spec\Http\Message;
```
dbu: why is this not in the Authentication namespace?

sagikazarmark: Good question. :smile: :+1: Should be there.

PR #5, opened 2015-12-20
Encoding

sagikazarmark:
see php-http/httplug#98

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,8 @@ branches:
         - /^analysis-.*$/
 
 matrix:
+    allow_failures:
+        - php: hhvm
     fast_finish: true
     include:
         - php: 5.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,6 @@ branches:
         - /^analysis-.*$/
 
 matrix:
-    allow_failures:
-        - php: hhvm
     fast_finish: true
     include:
         - php: 5.4
--- a/composer.json
+++ b/composer.json
@@ -13,19 +13,20 @@
     "require": {
         "php": ">=5.4",
         "psr/http-message": "^1.0",
-        "php-http/message-factory": "^1.0"
+        "php-http/message-factory": "^1.0",
+        "clue/stream-filter": "^1.3"
     },
     "require-dev": {
         "zendframework/zend-diactoros": "^1.0",
         "guzzlehttp/psr7": "^1.0",
         "ext-zlib": "*",
-        "clue/stream-filter": "^1.3",
         "phpspec/phpspec": "^2.4",
         "henrikbjorn/phpspec-code-coverage" : "^1.0"
     },
     "suggest": {
         "zendframework/zend-diactoros": "Used with Diactoros Factories",
-        "guzzlehttp/psr7": "Used with Guzzle PSR-7 Factories"
+        "guzzlehttp/psr7": "Used with Guzzle PSR-7 Factories",
+        "ext-zlib": "Used with compressor/decompressor streams"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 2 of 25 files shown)


.travis.yml
```diff
@@ -22,6 +22,8 @@ branches:
         - /^analysis-.*$/
 
 matrix:
+    allow_failures:
+        - php: hhvm
```
dbu: we could try to detect hhvm and trigger a skip just for the tests that don't work with hhvm rather than ignore hhvm issues over the whole package.

sagikazarmark: Good point.

composer.json
```diff
@@ -18,6 +18,8 @@
     "require-dev": {
         "zendframework/zend-diactoros": "^1.0",
         "guzzlehttp/psr7": "^1.0",
+        "ext-zlib": "*",
+        "clue/stream-filter": "^1.3",
```
joelwurtz: It should be in the require part, otherwise encoding does not work

sagikazarmark: Please see #8

dbu: lets add these to suggest, and explain for what case one would want to use them in a project.

src/Encoding/CompressStream.php
```diff
@@ -0,0 +1,28 @@
...
+class CompressStream extends FilteredStream
+{
+    public function __construct(StreamInterface $stream, $level = -1)
+    {
+        parent::__construct($stream, ['window' => 15, 'level' => $level], ['window' => 15]);
+    }
+
+    public function getReadFilter()
+    {
+        return 'zlib.deflate';
```
sagikazarmark: Should we throw an exception when zlib is not found, but this stream is being used?

dbu: good idea. with the optional dependency, this will at least explain what is going on. the exception message should say that you need the ext.zlib in your project if you want to use this.

sagikazarmark:
I had to add ignore platform deps hack to the build. As soon as we drop 5.4, it can be removed.

sagikazarmark:
I think this is ready to be merged.

.travis.yml
```diff
@@ -25,7 +24,9 @@ matrix:
     fast_finish: true
     include:
         - php: 5.4
-          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" COVERAGE=true TEST_COMMAND="composer test-ci"
+          env: COMPOSER_FLAGS="--ignore-platform-reqs"
```
dbu: is this because of the stream-filter library? if so i would move it to require-dev and suggest. if we support php 5.4, we should not fail with weird errors or tell people to ignore the platform and then they are surprised when trying to use streams.

for travis, we will need to keep the flag, as we do need the dev dependencies.

sagikazarmark: No, it is required for php-mock, which I used to mock `extension_loaded` function to test exceptions thrown.

sagikazarmark:
Replaced the php-mock dependency with a custom solution.

src/Encoding/InflateStream.php
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Stream inflate (RFC 1951).
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+class InflateStream extends FilteredStream
+{
+    public function __construct(StreamInterface $stream, $level = -1)
+    {
+        parent::__construct($stream, ['window' => -15], ['window' => -15, 'level' => $level]);
```
dbu: does this not need a check for zlib as well?

dbu:
apart from the zlib check on InflateStream, this looks ready to merge. not sure if you want to squash the commits or not.

sagikazarmark:
No squashing is needed IMO as these commits rather make sense to be preserved. I squashed some which were related to the zlib check.

PR #9, opened 2015-12-21
Add query param authentication

sagikazarmark:


dbu:
does this exist? it sounds like a rather dangerous way of doing things. if its ever done with a browser, it ends up in the history, and could even be spilled as a referrer. i would expect to use a post body in that case.

that said, if there are systems out there doing this, we can have it but i would like to have a little bit more documentation that warns that this is not a good idea in general.

sagikazarmark:
Unfortunately, this exists. Not just sort of acceptable solutions (like some request token), but user/password pairs as well (just did it connecting to a third party API, so from my point of view I need this feature, their security sucks :stuck_out_tongue_winking_eye: )

dbu:
oh wow. okay. but can you please update the phpdoc a bit, to explain
that this is not a good practice?

sagikazarmark:
Sure, will also add it to the documentation (php-http/documentation#43)

sagikazarmark:
@dbu is it better this way? I will also place a big fat warning in the docs.

PR #17, opened 2015-12-24
Formatter

sagikazarmark:
TODO:

- [ ] Implement cURL formatter


dbu:
:+1: - the curl formatter could also be done later in a separate PR

PR #22, opened 2015-12-27
Add cookie

### LINKED ISSUE #15: Cookie
Move and finalize php-http/cookie here.

sagikazarmark:
This PR adds Cookie implementation as discussed in php-http/httplug#98

Fix #15

A few open issues left in the original repository:
https://github.com/php-http/cookie/issues

- php-http/cookie#4: Make CookieJar a single class and support Persistence through a separate interface
- php-http/cookie#5: Double check matching implementation, I remember I had some issues with it

### DIFF
--- a/src/Cookie.php
+++ b/src/Cookie.php
@@ -1,0 +1,478 @@
+<?php
+
+namespace Http\Message;
+
+/**
+ * Cookie Value Object.
+ *
+ * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
+ *
+ * @see http://tools.ietf.org/search/rfc6265
+ */
+final class Cookie
+{
+    /**
+     * @var string
+     */
+    private $name;
+
+    /**
+     * @var string|null
+     */
+    private $value;
+
+    /**
+     * @var int|null
+     */
+    private $maxAge;
+
+    /**
+     * @var string|null
+     */
+    private $domain;
+
+    /**
+     * @var string
+     */
+    private $path;
+
+    /**
+     * @var bool
+     */
+    private $secure;
+
+    /**
+     * @var bool
+     */
+    private $httpOnly;
+
+    /**
+     * Expires attribute is HTTP 1.0 only and should be avoided.
+     *
+     * @var \DateTime|null
+     */
+    private $expires;
+
+    /**
+     * @param string         $name
+     * @param string|null    $value
+     * @param int            $maxAge
+     * @param string|null    $domain
+     * @param string|null    $path
+     * @param bool           $secure
+     * @param bool           $httpOnly
+     * @param \DateTime|null $expires  Expires attribute is HTTP 1.0 only and should be avoided.
+     *
+     * @throws \InvalidArgumentException If name, value or max age is not valid.
+     */
+    public function __construct(
+        $name,
+        $value = null,
+        $maxAge = null,
+        $domain = null,
+        $path = null,
+        $secure = false,
+        $httpOnly = false,
+        \DateTime $expires = null
+    ) {
+        $this->validateName($name);
+        $this->validateValue($value);
+        $this->validateMaxAge($maxAge);
+
+        $this->name = $name;
+        $this->value = $value;
+        $this->maxAge = $maxAge;
+        $this->expires = $expires;
+        $this->domain = $this->normalizeDomain($domain);
+        $this->path = $this->normalizePath($path);
+        $this->secure = (bool) $secure;
+        $this->httpOnly = (bool) $httpOnly;
+    }
+
+    /**
+     * Returns the name.
+     *
+     * @return string
+     */
+    public function getName()
+    {
+        return $this->name;
+    }
+
+    /**
+     * Returns the value.
+     *
+     * @return string|null
+     */
+    public function getValue()
+    {
+        return $this->value;
+    }
+
+    /**
+     * Checks if there is a value.
+     *
+     * @return bool
+     */
+    public function hasValue()
+    {
+        return isset($this->value);
+    }
+
+    /**
+     * Sets the value.
+     *
+     * @param string|null $value
+     *
+     * @return Cookie
+     */
+    public function withValue($value)
+    {
+        $this->validateValue($value);
+
+        $new = clone $this;
+        $new->value = $value;
+
+        return $new;
+    }
+
+    /**
+     * Returns the max age.
+     *
+     * @return int|null
+     */
+    public function getMaxAge()
+    {
+        return $this->maxAge;
+    }
+
+    /**
+     * Checks if there is a max age.
+     *
+     * @return bool
+     */
+    public function hasMaxAge()
+    {
+        return isset($this->maxAge);
+    }
+
+    /**
+     * Sets the max age.
+     *
+     * @param int|null $maxAge
+     *
+     * @return Cookie
+     */
+    public function withMaxAge($maxAge)
+    {
+        $this->validateMaxAge($maxAge);
+
+        $new = clone $this;
+        $new->maxAge = $maxAge;
+
+        return $new;
+    }
+
+    /**
+     * Returns the expiration time.
+     *
+     * @return \DateTime|null
+     */
+    public function getExpires()
+    {
+        return $this->expires;
+    }
+
+    /**
+     * Checks if there is an expiration time.
+     *
+     * @return bool
+     */
+    public function hasExpires()
+    {
+        return isset($this->expires);
+    }
+
+    /**
+     * Sets the expires.
+     *
+     * @param \DateTime|null $expires
+     *
+     * @return Cookie
+     */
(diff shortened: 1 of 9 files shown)


src/Cookie.php
```diff
@@ -0,0 +1,494 @@
...
+    }
+
+    /**
+     * Sets both the max age and the expires attributes.
+     *
+     * @param int|\DateTime|null $expiration
+     *
+     * @return self
+     */
+    public function withExpiration($expiration)
```
dbu: what is the ordering logic of these methods? for max-age its get/has/with, here its with/get/has. i propose we sort all of them the same way. has/get/with feels most natural to me but i am happy with any, as long as its consistent.

sagikazarmark: That's just a mistake. I usually follow get/has/set order.

dbu: ah sorry, i whould have read the the phpdoc. but then i don't get what the idea is. if the param is an int, we use it as max-age, if its a datetime we use it as expires? are we sure that this is a good idea?

i would prefer to have withMaxAge remove the expires property, and vice versa.

dbu: expires is the old way of doing things, HTTP 1.1 has max-age: http://mrcoles.com/blog/cookies-max-age-vs-expires/ (IE9+ supports max-age). i guess we can keep expires support, but i would not support the mixed approach. if anything, it would be the job of some hypothetical server side code that sets expires from the max-age.

though when we speak about client side code, we could receive a cookie with both values set and might care about that (and be it just to report that the server is sending ambigous data) - so probably setting one value should not null out the other. 

but i would still remove this method, it seems weird. and we could phpdoc that expires is the old way of doing things and no longer recomended

src/Cookie.php
```diff
@@ -0,0 +1,494 @@
...
+    public function withDomain($domain)
+    {
+        $new = clone $this;
+        $new->domain = $this->normalizeDomain($domain);
+
+        return $new;
+    }
+
+    /**
+     * Matches a domain.
```
dbu: this comment says nothing. lets say: "Check whether this cookie is meant for this domain."

src/Cookie.php
```diff
@@ -0,0 +1,494 @@
...
+     *
+     * @param string $path
+     *
+     * @return bool
+     *
+     * @see http://tools.ietf.org/html/rfc6265#section-5.1.4
+     */
+    public function matchPath($path)
+    {
+        return $this->path === $path || (strpos($path, $this->path.'/') === 0);
```
dbu: are we sure about adding the trailing `/`?

src/Cookie.php
```diff
@@ -0,0 +1,494 @@
...
+    public function withPath($path)
+    {
+        $new = clone $this;
+        $new->path = $this->normalizePath($path);
+
+        return $new;
+    }
+
+    /**
+     * It matches a path.
```
dbu: "Check whether this cookie is meant for this path."

and maybe also mention matchDomain as the second criteria to decide whether this cookie is valid for this request.

src/Cookie.php
```diff
@@ -0,0 +1,494 @@
...
+     *
+     * @return bool
+     */
+    public function isSecure()
+    {
+        return $this->secure;
+    }
+
+    /**
+     * Sets the secure.
```
dbu: Sets whether this cookie should only be sent over HTTPS.

src/Cookie.php
```diff
@@ -0,0 +1,494 @@
...
+     *
+     * @see http://tools.ietf.org/html/rfc6265#section-5.1.4
+     */
+    public function matchPath($path)
+    {
+        return $this->path === $path || (strpos($path, $this->path.'/') === 0);
+    }
+
+    /**
+     * Checks if HTTPS is required.
```
dbu: Checks whether this cookie may only be sent over HTTPS.

src/Cookie.php
```diff
@@ -0,0 +1,494 @@
...
+    public function withSecure($secure)
+    {
+        $new = clone $this;
+        $new->secure = (bool) $secure;
+
+        return $new;
+    }
+
+    /**
+     * Checks if it is HTTP-only.
```
dbu: Check whether this cookie may not be accessed through Javascript.

src/Cookie.php
```diff
@@ -0,0 +1,494 @@
...
+     *
+     * @return bool
+     */
+    public function isHttpOnly()
+    {
+        return $this->httpOnly;
+    }
+
+    /**
+     * Sets the HTTP Only.
```
dbu: Sets whether this cookie may not be accessed through Javascript.

src/Cookie.php
```diff
@@ -0,0 +1,494 @@
...
+    public function withHttpOnly($httpOnly)
+    {
+        $new = clone $this;
+        $new->httpOnly = (bool) $httpOnly;
+
+        return $new;
+    }
+
+    /**
+     * Checks if it matches with another cookie.
```
dbu: Checks if this cookie represents the same cookie as $cookie.

This does not compare the values, only name, domain and path.

src/CookieJar/PersistentCookieJar.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace Http\Message\CookieJar;
+
+use Http\Message\CookieJar;
+
+/**
+ * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
+ */
+interface PersistentCookieJar extends CookieJar
```
dbu: is this our job or should that be handled with instantiation? it seems a bit out of scope here to define this.

src/Cookie.php
joelwurtz: :+1:

src/CookieJar/PersistentCookieJar.php
sagikazarmark: Nope, see php-http/cookie#4

sagikazarmark:
Thanks for the feedback.

src/Cookie.php
sagikazarmark: Nope, see php-http/cookie#5

sagikazarmark:
About the expires thing: currently you can pass an expiration parameter in the constructor. Should we have max age instead. Also, this expiration is set for both max age and expires property ATM.

Should we keep both max age and expires and only remove the magic expiration method?

dbu:
i would allow to set both from the constructor, to make it possible to
model a response precisely.

but remove the magic method.

and phpdoc that expire is http 1.0 and should be avoided.

sagikazarmark:
What about order of arguments: should expires be the last?

src/Cookie.php
```diff
@@ -0,0 +1,478 @@
...
+     * @param bool           $secure
+     * @param bool           $httpOnly
+     * @param \DateTime|null $expires
+     *
+     * @throws \InvalidArgumentException
+     */
+    public function __construct(
+        $name,
+        $value = null,
+        $maxAge = 0,
```
dbu: is this correct or should default value be null?

dbu:
:+1: agree to have expires last so it can be ignored when not recording an exact Server response

src/Cookie.php
```diff
@@ -0,0 +1,478 @@
...
+
+    /**
+     * @param string         $name
+     * @param string|null    $value
+     * @param int            $maxAge
+     * @param string|null    $domain
+     * @param string|null    $path
+     * @param bool           $secure
+     * @param bool           $httpOnly
+     * @param \DateTime|null $expires
```
dbu: lets add the deprication note here too

sagikazarmark: Should we add a `@deprecated` phpdoc to methods as well?

`@deprecated since HTTP 1.0`

src/Cookie.php
sagikazarmark: Checked it. Yes, the trailing slash is necessary.

http://tools.ietf.org/html/rfc6265#section-5.1.4
http://tools.ietf.org/html/rfc6265#section-5.2.4

src/Cookie.php
dbu: > Should we add a @deprecated phpdoc to methods as well?

i would not do that. its not like we will remove the method.

sagikazarmark:
We could add cookie related helpers here.

One thing I am thinking about:

We could have the cookie parser as is OR provide a compat package around this method:

http://php.net/manual/en/function.http-parse-cookie.php

In general, we could have a pecl_http compatibility layer, but I have never used it before and don't plan in the future, so I wonder if it makes sense.

sagikazarmark:
Let's make cookie parsers in another PR. I think this is good to merge. Checked RFC compliance, but I am not sure about it. I've read the RFC several times, it is not always clear. From what I can understand, everything is OK.

dbu:
the pecl_http compatibility layer sounds like a good idea but lets move forward. it can always be added later.

PR #24, opened 2015-12-31
Require message-factory version that has Puli bindings

ddeboer:
message-factory 1.0.0 didn't yet have Puli bindings, which causes trouble
when installing php-http packages using Composer's --prefer-lowest.


sagikazarmark:
:+1:

PR #26, opened 2016-01-25
Lower buffer size, and use it in the getContents

joelwurtz:
This is mainly to avoid too much nested call to the read method, as if we got a stream where the reads provide too few bytes (like 100) it will make a nested call to the read method for each part of the messages and will stop on a too muche nested call exception with xdebug.

Also there is no need to have more than 8192 as PHP already have a buffer with this size, so no performance issue.


dbu:
:+1:

PR #30, opened 2016-01-29
ResponseBuilder class.

mekras:
client-tools repo was removed, so I recreated PR for ResponseBuilder here.

Not sure about spec file because I'm not familiar with PhpSpec.

### DIFF
--- a/spec/Builder/ResponseBuilderSpec.php
+++ b/spec/Builder/ResponseBuilderSpec.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace spec\Http\Message\Builder;
+
+use PhpSpec\ObjectBehavior;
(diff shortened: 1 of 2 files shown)


joelwurtz:
Don't really understand the purpose of such builder ? Why not using a response factory or directly the response ?

mekras:
@joelwurtz, we have response factory which can do this?

The purpose is reuse of implementation independent code. It was discussed in original PR. Parsing headers is a common task for HTTP client (similar code in socket and curl clients). Also it can be used by third-party libs. For example I need to create HttpMessage instances from HTTP dumps and so can use this functionality.

joelwurtz:
Hum didn't read well the code, my bad ^^

LGTM :+1:

PR #32, opened 2016-02-19
Add request matcher

joelwurtz:
See https://github.com/php-http/client-common/pull/13

### DIFF
--- a/src/RequestMatcher.php
+++ b/src/RequestMatcher.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace Http\Message;
+
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Match a request.
+ *
+ * PSR-7 equivalent of Symfony's RequestMatcher
+ *
+ * @see https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/RequestMatcherInterface.php
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+interface RequestMatcher
+{
+    /**
+     * Decides whether the rule(s) implemented by the strategy matches the supplied request.
+     *
+     * @param RequestInterface $request The PSR7 request to check for a match
+     *
+     * @return bool true if the request matches, false otherwise
+     */
+    public function matches(RequestInterface $request);
+}
--- a/src/RequestMatcher/RegexRequestMatcher.php
+++ b/src/RequestMatcher/RegexRequestMatcher.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Http\Message\RequestMatcher;
+
+use Http\Message\RequestMatcher;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Match a request with a regex on the uri.
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+class RegexRequestMatcher implements RequestMatcher
+{
+    /**
+     * Matching regex.
+     *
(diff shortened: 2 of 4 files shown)


src/RequestMatcher.php
```diff
@@ -0,0 +1,25 @@
...
+ * Match a request.
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+interface RequestMatcher
+{
+    /**
+     * Decides whether the rule(s) implemented by the strategy matches the supplied request.
+     *
+     * Equivalent of https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/RequestMatcherInterface.php
```
sagikazarmark:
```
PSR-7 implementation of Symfony's RequestMatcher

@see https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/RequestMatcherInterface.php
```

src/RequestMatcher/RegexRequestMatcher.php
```diff
@@ -0,0 +1,30 @@
...
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Match a request with a regex on the uri.
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+class RegexRequestMatcher implements RequestMatcher
+{
+    /** @var string Matching regex */
```
sagikazarmark: Please expand this to multiline.

sagikazarmark:
Other possible matchers to add:

- CallbackMatcher
- ChainMatcher (AND or OR logic?)

sagikazarmark:
LGTM :+1:

joelwurtz:
Other matchers for another PR when we need it :)

dbu:
i like the symfony approach to this better, actually: https://github.com/symfony/symfony/blob/582f4753a343f230fbe18b4e9a0747d48351ddfb/src/Symfony/Component/HttpFoundation/RequestMatcher.php

if we have a separate matcher for each part of the request, we build huge object trees.
i don't know if anybody ported the symfony request matcher to PSR-7 yet, if not this library could be a good place for it. though it might be worth to have it stand alone as it could be useful even when not using anything else from PHP-HTTP.

sagikazarmark:
I don't see a reason why the regex one could not live together with the one you linked.

The message library is not dedicated to PHP-HTTP packages (but rather to PSR-7 as a utility package), so IMO we can have it in here. There aren't any heavy dependencies. (BTW: We should ask the FIG if they want a http-message-utils repo, similarly to cache-utils.)

But I am not against separating it to be standalone. (One reason against it: makes it harder to use it INSIDE message. For example I would use it the Matching authentication)

dbu:
agreed, lets keep it inside message then. good idea to ask fig - if we
contribute parts of this to their world we can then depend on those
parts from message without worries and still rely on it.

my unhappiness with the RegexRequestMatcher is that it is only matching
one thing, instead of ip, host, path, method and so on like the symfony
one. if we add the full matcher, the regex matcher becomes redundant.
the name is a bit unclear too, its a RegexUriRequestMatcher, not regex
for host, path or method...

joelwurtz:
> my unhappiness with the RegexRequestMatcher is that it is only matching
one thing

It's just a first implementation to have a simple example, as i'm only interested in the interface for the moment, will certainly write other implementations at a latter time :)

It can be renamed if you want, also didn't want a too long name, but yeah RegexUriRequestMatcher may be cleaner.

PR #35, opened 2016-03-02
Add new RequestMatch authentication method

dbu:
@sagikazarmark found this branch lying around. it seems to add on #34 - is it relevant?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,12 @@
 
 ## Unreleased
 
+### Added
 
-## 1.1.0
+- New RequestConditional authentication method using request matchers (deprecates Matching auth method)
+
+
+## 1.1.0 - 2016-02-25
 
 ### Added
 
@@ -13,7 +17,8 @@
 
 ### Fixed
 
- - Fix casting string on a FilteredStream not filtering the output 
+ - Fix casting string on a FilteredStream not filtering the output
+
 
 ## 1.0.0 - 2016-01-27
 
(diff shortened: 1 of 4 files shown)


src/Authentication/RequestMatch.php
```diff
@@ -0,0 +1,47 @@
...
+use Http\Message\Authentication;
+use Http\Message\RequestMatcher;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Authenticate a PSR-7 Request if the request is matching the given request matcher.
+ *
+ * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
+ */
+final class RequestMatch implements Authentication
```
dbu: maybe we could call this ConditionalAuthentication?

sagikazarmark: We should somehow have the request in the name...otherwise what condition would it be?

sagikazarmark:
Yeah, I replaced the matching auth with the request matcher interface one. It indeed relied on that PR, as the deprecated matcher now uses the callback request matcher internally.

src/Authentication/RequestMatch.php
sagikazarmark: RequestConditional?

sagikazarmark: RequestAware?

sagikazarmark: I fear we can't find a self-explanatory name enough, we have to document it anyway.

dbu: i would like RequestConditional better. and explain it - the phpdoc above looks fine to me.

PR #36, opened 2016-03-02
adjust request matcher to the symfony model

### LINKED ISSUE #33: Replace regex/add symfony-like request matcher
See #32 

https://github.com/symfony/symfony/blob/582f4753a343f230fbe18b4e9a0747d48351ddfb/src/Symfony/Component/HttpFoundation/RequestMatcher.php

dbu:
fix #33

this matcher has less options than the Symfony request matcher. a Symfony request is more like the ServerRequestInterface of psr 7 and we mainly work with client requests, so matching on the client ip makes no sense. and psr 7 requests have no attributes so i dropped that as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Added
 
+- The RequestMatcher is built after the Symfony RequestMatcher and separates
+   scheme, host and path expressions and provides an option to filter on the
+   method. The RegexRequestMatcher is deprecated and will be removed in 2.0.
 - New RequestConditional authentication method using request matchers (deprecates Matching auth method)
 
 
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.1-dev"
+            "dev-master": "1.2-dev"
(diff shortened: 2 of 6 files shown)


sagikazarmark:
Isn't this a BC break? I think that message has already been tagged by @joelwurtz

dbu:
code using the matcher would still "work" but the meaning of the first parameter has changed. it used to match anything of the uri: scheme, host or path. now it only applies to path. not sure if that could be a 1.2 version or is really too bad a BC break.

not sure how we should handle this. i don't want to add a separate class that does the same but different. should we delay merging this until we bump to 2.0?

sagikazarmark:
Well, we either add a separate one which does the same and deprecates the other (and gets removed in 2.0) or just do this and break BC in 1.2. That would require a big fat warning.

From a BC point of view the first one seems to be better to me. Although I am not sure how many people might already use this feature (but this mentality is not the best for versioning).

dbu:
okay, you are right. we do semver, we should not randomly decide to violate BC.

deprecated and add a new one as RequestMatcher. what do you think of this?

sagikazarmark:
Sounds like a plan

PR #38, opened 2016-03-04
Add auto BasicAuth credential detection based on URL

sagikazarmark:


xabbuh:
Could this work with Digest authentication too?

sagikazarmark:
Good question.

I think as in case of oauth, digest auth can only work as a separate plugin as it needs a request first in order to retrieve the server nonce.

So with a slightly difference code and a different plugin: probably yes.

But as a plain authentication method: not sure. Since the method is immutable, you have to provide every data in the constructor.

spec/Authentication/AuthenticationBehavior.php
```diff
@@ -1,11 +0,0 @@
-<?php
-
-namespace spec\Http\Message\Authentication;
-
-trait AuthenticationBehavior
```
dbu: did this not work as expected?

dbu:
is this what a web browser does with such urls? then i agree this is a good thing to have. is there a corresponding doc PR already? if not, at least create an issue - if we lose track of features the doc will quickly go out of sync and it becomes a huge pain to figure out discrepancies (been there, done that)

spec/Authentication/AuthenticationBehavior.php
sagikazarmark: It did (partly). However PHPSpec currently does not support more complex examples (eg. it cannot use prophecy as function arguments) because of some sort of autoloading issues. So I decided to remove this because I though it didn't make sense just to avoid duplication of one example.

dbu: if there is a technical problem with it, i agree its ok to duplicate
that method.

sagikazarmark:
Yes, this is a client responsibility. Will open a doc PR. It will probably be easier if we start using issue/PR templates: https://github.com/php-http/boilerplate/tree/master/.github

PR #40, opened 2016-03-23
Single header line parsing extracted to a separate method in ResponseBuilder

mekras:
Allows more flexible use of the class.

### DIFF
--- a/src/Builder/ResponseBuilder.php
+++ b/src/Builder/ResponseBuilder.php
@@ -48,38 +48,13 @@ class ResponseBuilder
      */
     public function setHeadersFromArray(array $headers)
     {
-        $statusLine = trim(array_shift($headers));
-        $parts = explode(' ', $statusLine, 3);
-        if (count($parts) < 2 || substr(strtolower($parts[0]), 0, 5) !== 'http/') {
-            throw new \UnexpectedValueException(
-                sprintf('"%s" is not a valid HTTP status line', $statusLine)
-            );
-        }
-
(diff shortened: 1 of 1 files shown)


sagikazarmark:
LGTM. Wonder if we should add more tests for the addHeader method separately.

mekras:
@sagikazarmark, may be. But a) new method is fully covered with existed tests; b) 
I don't feel familiar enough with phpspec.

sagikazarmark:
Ok. Lemme check if I can do something on this front during the weekend.

src/Builder/ResponseBuilder.php
```diff
@@ -48,38 +48,13 @@ public function getResponse()
...
-            throw new \UnexpectedValueException(
-                sprintf('"%s" is not a valid HTTP status line', $statusLine)
-            );
-        }
-
-        $reasonPhrase = count($parts) > 2 ? $parts[2] : '';
-        $this->response = $this->response
-            ->withStatus((int) $parts[1], $reasonPhrase)
-            ->withProtocolVersion(substr($parts[0], 5));
-
```
joelwurtz: I dont think this should be handled via the addHeaderLine as it's not a header but more the status string, also now a request like this : 

```
Header: headervalue
GET / HTTP/1.1
```

would be correctly parsed but his invalid and should throw an error

mekras: Changed.

joelwurtz:
:+1: LGTM

PR #42, opened 2016-06-24
Fix #41: Response builder broke header value

### LINKED ISSUE #41: Response builder broke header value
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 1.2.0

#### Actual Behavior

Curl client (and maybe more) using `Http\Message\Builder\ResponseBuilder`. When client adds `Content-Type: application/hal+json` header using `addHeader` method, created response contains `Content-Type: application/hal json`

#### Expected Behavior

Created response should contains valid header value as it's in raw response.

#### Steps to Reproduce

Send request to get response with header contains `+` eg. `Content-Type: application/hal+json`

#### Possible Solutions

Remove / fix decoding in https://github.com/php-http/message/blob/master/src/Builder/ResponseBuilder.php#L139

mekras:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #41 

#### Checklist

### DIFF
--- a/spec/Builder/ResponseBuilderSpec.php
+++ b/spec/Builder/ResponseBuilderSpec.php
@@ -32,4 +32,17 @@ class ResponseBuilderSpec extends ObjectBehavior
         $this->beConstructedWith($response);
         $this->setHeadersFromString("HTTP/1.1 200 OK\r\nContent-type: text/html\r\n");
     }
+
+    /**
+     * @link https://github.com/php-http/message/issues/41
+     */
+    function it_not_broke_headers(ResponseInterface $response)
+    {
+        $response->withStatus(200, 'OK')->willReturn($response);
+        $response->withProtocolVersion('1.1')->willReturn($response);
+        $response->hasHeader('Content-type')->willReturn(false);
+        $response->withHeader('Content-type', 'application/xml+atom')->willReturn($response);
+        $this->beConstructedWith($response);
+        $this->setHeadersFromString("HTTP/1.1 200 OK\r\nContent-type: application/xml+atom\r\n");
+    }
 }
--- a/spec/Builder/ResponseBuilderSpec.php
+++ b/spec/Builder/ResponseBuilderSpec.php
@@ -23,20 +23,10 @@ class ResponseBuilderSpec extends ObjectBehavior
         $this->setHeadersFromArray(['HTTP/1.1 200 OK', 'Content-type: text/html']);
     }
 
-    function it_reads_headers_from_string(ResponseInterface $response)
-    {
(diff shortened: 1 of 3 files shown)


src/Builder/ResponseBuilder.php
```diff
@@ -135,8 +135,8 @@ public function addHeader($headerLine)
                 sprintf('"%s" is not a valid HTTP header line', $headerLine)
             );
         }
-        $name = trim(urldecode($parts[0]));
-        $value = trim(urldecode($parts[1]));
+        $name = trim($parts[0]);
+        $value = trim($parts[1]);
```
dbu: the urldecode was in there from the beginning. is something else doing that decoding in reality? is your test failing if you keep the urldecode here?

spec/Builder/ResponseBuilderSpec.php
```diff
@@ -32,4 +32,17 @@ function it_reads_headers_from_string(ResponseInterface $response)
         $this->beConstructedWith($response);
         $this->setHeadersFromString("HTTP/1.1 200 OK\r\nContent-type: text/html\r\n");
     }
+
+    /**
+     * @link https://github.com/php-http/message/issues/41
+     */
+    function it_not_broke_headers(ResponseInterface $response)
```
dbu: maybe call this `it_splits_headers_correctly`?

src/Builder/ResponseBuilder.php
mekras: > the urldecode was in there from the beginning. is something else doing that decoding in reality?

See comment for https://github.com/php-http/message/pull/42/commits/60fe987ed889ce4b0d9c1c7a81a4e81153b5ee4b

> is your test failing if you keep the urldecode here?

Yes.

spec/Builder/ResponseBuilderSpec.php
mekras: I don't care.

src/Builder/ResponseBuilder.php
sagikazarmark: It seems to me the urldecode is not necessary, only ascii characters are allowed.

spec/Builder/ResponseBuilderSpec.php
sagikazarmark: Please do the rename @dbu suggested as it better explains what we exactly test there.

sagikazarmark:
Thank you for your work @mekras. Looks good to me. I am happy to merge it after a squash.

dbu:
thanks!

PR #44, opened 2016-07-03
Added a formatter that shows the full header and body of the HTTP message

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| Documentation   | no
| License         | MIT

#### What's in this PR?

A new formatter that show the full HTTP message. The formatter takes a constructor argument to adjust the length of the body. If the length is set to `0` we would only see the headers. (naturally). 

#### Why?

I will need this formatter in a (WIP) PR on the HTTPlug bundle. I want to show the changes each Plugin does on the request and responses. 

Im not sure this formatter belongs here, but I think there is a fairly common use case to be able to see full requests. If it is rejected here I will push it to the HTTPlug repo. 

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,13 @@
 
 ## Unreleased
 
-### Fixed
- 
-- #41: Response builder broke header value
-
 ### Added
 
 - The FullHttpMessageFormatter was added
+
+### Fixed
+ 
+- #41: Response builder broke header value
 
 ## 1.2.0 - 2016-03-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
  
 - #41: Response builder broke header value
 
+### Added
+
+- The FullHttpMessageFormatter was added
 
 ## 1.2.0 - 2016-03-29
 
--- a/src/Formatter/FullHttpMessageFormatter.php
+++ b/src/Formatter/FullHttpMessageFormatter.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Http\Message\Formatter;
+
+use Http\Message\Formatter;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * A formatter that prints the complete HTTP message.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class FullHttpMessageFormatter implements Formatter
+{
+    /**
+     * The maximum length of the body.
+     *
+     * @var int
+     */
+    private $maxBodyLendth;
+
+    /**
+     * @param int $maxBodyLendth number of chars.
+     */
+    public function __construct($maxBodyLendth = 1000)
+    {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
  
 - #41: Response builder broke header value
 
+### Added
```
sagikazarmark: Please keep alphabetic order in the changelog.

src/Formatter/FullHttpMessageFormatter.php
```diff
@@ -0,0 +1,72 @@
...
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class FullHttpMessageFormatter implements Formatter
+{
+    /**
+     * The maximum length of the body.
+     *
+     * @var int
+     */
+    private $maxBodyLendth;
```
sagikazarmark: s/Lendth/Length/

src/Formatter/FullHttpMessageFormatter.php
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * @param int $maxBodyLendth number of chars.
```
sagikazarmark: Description seems to be redundant. IMO the property description is enough.

src/Formatter/FullHttpMessageFormatter.php
```diff
@@ -0,0 +1,72 @@
...
+            $request->getMethod(),
+            $request->getRequestTarget(),
+            $request->getProtocolVersion()
+        );
+
+        foreach ($request->getHeaders() as $name => $values) {
+            $message .= $name.': '.implode(', ', $values)."\n";
+        }
+
+        $message .= "\n".mb_substr($request->getBody()->__toString(), 0, $this->maxBodyLendth);
```
sagikazarmark: Wonder if string casting or calling toString is better. Anyone knows any recommendation for that. IMO it is a little bit weird that __toString was included in PSR-7 interfaces.

sagikazarmark:
LGTM, added a few comments.

src/Formatter/FullHttpMessageFormatter.php
Nyholm: What is the difference? Sure one may think one **looks** better than the other. But is one faster then the other? String casting of an object calls __toString() as far as I know. 

I like calling __toString but Im happy to change if the other way is better or more coherent with the rest of our code.

sagikazarmark: Yes, they are the same. I am just looking for a concept we should follow.

dbu:
thanks tobias, you had a busy weekend!

we might expand this to have a limit on each header as well. headers can technically be very long as well, at least up to 1000 characters. having a limit of 100 or 200 would rarely be hit, but if it is hit it would be relevant to not drown a log in gibberish... but this can be added later if somebody wants to make the effort.

sagikazarmark:
Yeah, if we had an MVP, that would be @Nyholm now.

PR #47, opened 2016-07-15
Rewind streams

### LINKED ISSUE #46: Rewind streams in the formatter
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 1.3.0

#### Actual Behavior

From: https://github.com/php-http/HttplugBundle/pull/84#issuecomment-232911288
> Just see the formatter and how the request / response is read.

>As far as i have see, this formatter mess up with the stream, as __toString is called and the stream is not rewind so the content is retrieve only in the first stream ?

> (I may have miss something again ^^)

#### Expected Behavior

Streams should be rewinded after they are read.

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | ~~yes~~ no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #46
| Documentation   | n/a
| License         | MIT

#### What's in this PR?

* We rewind streams if we can. 
* We do not read streams if maxBodyLength is 0
* We do not read streams we can rewind.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## 1.3.1 - 2016-07-15
+
+### Fixed
+
+- FullHttpMessageFormatter will not read from streams that you cannot rewind (non-seekable)
+- FullHttpMessageFormatter will not read from the stream if $maxBodyLength is zero
+- FullHttpMessageFormatter rewinds streams after they are read.
 
 ## 1.3.0 - 2016-07-14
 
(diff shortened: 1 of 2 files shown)


dbu:
looks good to me! i would consider this a bugfix btw, not a new feature. master already has its branch-alias at 1.4 but no commits since 1.3 so we can simply tag 1.3.1 on master once this is merged.

Nyholm:
Agreed. This is a bugfix.

sagikazarmark:
Can you please add changelog for this?

Nyholm:
Of course!

Nyholm:
Thank you for the quick review.

PR #49, opened 2016-07-22
Add Message cloner

joelwurtz:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | mentioned in #46
| Documentation   | Does the phpdoc is enough ?
| License         | MIT

#### What's in this PR?

Add a class to copy a request including it's body

#### Why?

IMO cloning a request with its stream his something that is shared by some concepts (like logging, caching, ....)

#### Example Usage

```php
$cloner = new MessageCloner();

$clonedResponse = $cloner->cloneMessage($response);

// If original stream is not seekable, cloned it a second time to ensure that we can read it
if (!$response->getBody()->isSeekable()) {
    $response = $cloner->cloneMessage($clonedResponse);
}
```

#### Checklist

- [ ] Updated CHANGELOG.md to describe new feature
- [ ] Documentation pull request created
- [ ] Add a plugin "EnsureSeekable" : Goal is to have a plugin that ensure that the body of a request or response is always seekable.

### DIFF
--- a/spec/MessageClonerSpec.php
+++ b/spec/MessageClonerSpec.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace spec\Http\Message;
+
+use Http\Message\MemoryClonedStream;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use Psr\Http\Message\MessageInterface;
+use spec\Http\Message\Encoding\MemoryStream;
+
+class MessageClonerSpec extends ObjectBehavior
+{
+    public function it_is_initializable()
+    {
+        $this->shouldHaveType('Http\Message\MessageCloner');
+    }
+
+    public function it_clone_a_message(MessageInterface $message, MessageInterface $messageCloned)
+    {
+        $stream = new MemoryStream('test');
+        $message->getBody()->willReturn($stream);
+        $message->withBody(Argument::type(MemoryClonedStream::class))->willReturn($messageCloned);
+
+        $this->cloneMessage($message)->shouldEqual($messageCloned);
+    }
+}
--- a/spec/MessageClonerSpec.php
+++ b/spec/MessageClonerSpec.php
@@ -2,7 +2,6 @@
 
 namespace spec\Http\Message;
 
-use Http\Message\MemoryClonedStream;
 use PhpSpec\ObjectBehavior;
 use Prophecy\Argument;
 use Psr\Http\Message\MessageInterface;
@@ -19,7 +18,7 @@ class MessageClonerSpec extends ObjectBehavior
     {
         $stream = new MemoryStream('test');
         $message->getBody()->willReturn($stream);
-        $message->withBody(Argument::type(MemoryClonedStream::class))->willReturn($messageCloned);
+        $message->withBody(Argument::type('Http\Message\MemoryClonedStream'))->willReturn($messageCloned);
 
         $this->cloneMessage($message)->shouldEqual($messageCloned);
     }
--- a/src/MemoryClonedStream.php
+++ b/src/MemoryClonedStream.php
@@ -1,0 +1,182 @@
+<?php
+
+namespace Http\Message;
+
+use Psr\Http\Message\StreamInterface;
+
+/**
+ * Represent a stream cloned into memory.
+ */
+class MemoryClonedStream implements StreamInterface
+{
+    const COPY_BUFFER = 8192;
+
+    private $resource;
+
+    private $size;
+
+    /**
+     * @param StreamInterface $stream        Stream to clone
+     * @param bool            $useFileBuffer Use a file buffer to avoid memory consumption on PHP script (default to true)
+     * @param int             $memoryBuffer  The amount of memory of which the stream is buffer into a file when setting
+     *                                       $useFileBuffer to true (default to 2MB)
+     */
+    public function __construct(StreamInterface $stream, $useFileBuffer = true, $memoryBuffer = 2097152)
+    {
+        $this->size = 0;
+
+        if ($useFileBuffer) {
+            $this->resource = fopen('php://temp/maxmemory:'.$memoryBuffer, 'rw+');
+        } else {
+            $this->resource = fopen('php://memory', 'rw+');
+        }
+
+        $position = null;
+
+        if ($stream->isSeekable()) {
+            $position = $stream->tell();
+            $stream->rewind();
+        }
+
+        while (!$stream->eof()) {
+            $this->size += fwrite($this->resource, $stream->read(self::COPY_BUFFER));
+        }
+
+        if ($stream->isSeekable()) {
+            $stream->seek($position);
+        }
+
+        rewind($this->resource);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function __toString()
+    {
+        return $this->getContents();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function close()
+    {
+        fclose($this->resource);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function detach()
+    {
+        $resource = $this->resource;
+        $this->resource = null;
+
+        return $resource;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSize()
+    {
+        return $this->size;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function tell()
+    {
+        return ftell($this->resource);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function eof()
+    {
+        return feof($this->resource);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isSeekable()
+    {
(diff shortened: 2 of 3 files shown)


src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+use Psr\Http\Message\StreamInterface;
+
+/**
+ * Represent a stream cloned into memory.
+ */
+class MemoryClonedStream implements StreamInterface
+{
+    const COPY_BUFFER = 8192;
+
+    private $resource;
```
sagikazarmark: Can you add type hints here too for internal usage?

spec/MessageClonerSpec.php
```diff
@@ -0,0 +1,25 @@
...
+use spec\Http\Message\Encoding\MemoryStream;
+
+class MessageClonerSpec extends ObjectBehavior
+{
+    public function it_is_initializable()
+    {
+        $this->shouldHaveType('Http\Message\MessageCloner');
+    }
+
+    public function it_clone_a_message(MessageInterface $message, MessageInterface $messageCloned)
```
sagikazarmark: clones

sagikazarmark: $clonedMessage

sagikazarmark:
What about memory usage in case of huge bodies? We should add a warning somewhere about that.

joelwurtz:
It's in the header of the MemoryClonedStream, by default it use a php://temp stream with a 2mb memory buffer, so the memory footprint of a cloned stream will, by default, have a max memory size of 2mb, when the body is higher, all data over the limit is written into a file.

Nyholm:
I like this. It could help on all the places we do some ugly hack to rewind streams and make sure they are seekable. 

@joelwurtz could you address the comments?

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+
+    private $size;
```
Nyholm: Same here

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+
+        rewind($this->resource);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function __toString()
+    {
+        return $this->getContents();
```
Nyholm: You should rewind it first

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function seek($offset, $whence = SEEK_SET)
+    {
+        return fseek($this->resource, $offset);
```
Nyholm: I think psr7 says you should throw exception of whence is not supported.

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        return fread($this->resource, $length);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getContents()
+    {
+        $this->rewind();
```
Nyholm: PSR7 says that getContents should return "the rest of the string" if I remember corectly. Do not rewind here

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+     */
+    public function getMetadata($key = null)
+    {
+        $metadata = stream_get_meta_data($this->resource);
+
+        if (null === $key) {
+            return $metadata;
+        }
+
+        return $metadata[$key];
```
Nyholm: Verify with isset

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function detach()
+    {
+        $resource = $this->resource;
+        $this->resource = null;
+
+        return $resource;
```
Nyholm: Size should be set to zero

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+            $this->size += fwrite($this->resource, $stream->read(self::COPY_BUFFER));
+        }
+
+        if ($stream->isSeekable()) {
+            $stream->seek($position);
+        }
+
+        rewind($this->resource);
+    }
+
```
Nyholm: Add __destruct
```php
public function __destruct()
{
    $this->close();
}
```

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        return $this->getContents();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function close()
+    {
+        fclose($this->resource);
```
Nyholm: Check if `$this->resource` is not null.

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+    public function __construct(StreamInterface $stream, $useFileBuffer = true, $memoryBuffer = 2097152)
+    {
+        $this->size = 0;
+
+        if ($useFileBuffer) {
+            $this->resource = fopen('php://temp/maxmemory:'.$memoryBuffer, 'rw+');
+        } else {
+            $this->resource = fopen('php://memory', 'rw+');
+        }
+
```
Nyholm: Maybe verify that $this->resource is not false.

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        return $this->size;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function tell()
+    {
+        return ftell($this->resource);
```
Nyholm: Check if `$this->resource` is not null.

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function eof()
+    {
+        return feof($this->resource);
```
Nyholm: Check if `$this->resource` is not null.

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        return fseek($this->resource, $offset);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function rewind()
+    {
+        return rewind($this->resource);
```
Nyholm: Check if `$this->resource` is not null.

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function write($string)
+    {
+        return fwrite($this->resource, $string);
```
Nyholm: Check if `$this->resource` is not null.

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function read($length)
+    {
+        return fread($this->resource, $length);
```
Nyholm: Check if `$this->resource` is not null.

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        return feof($this->resource);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isSeekable()
+    {
+        return true;
```
Nyholm: Return false if $this->resource is null.

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        return rewind($this->resource);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isWritable()
+    {
+        return true;
```
Nyholm: Return false if $this->resource is null.

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        return fwrite($this->resource, $string);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isReadable()
+    {
+        return true;
```
Nyholm: Return false if $this->resource is null.

src/MemoryClonedStream.php
```diff
@@ -0,0 +1,182 @@
...
+
+        return $this->read($this->size);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getMetadata($key = null)
+    {
+        $metadata = stream_get_meta_data($this->resource);
```
Nyholm: Check if $this->resource is not null.

joelwurtz:
Finally i'm not a fan of this implementation, message cloner is useless and the cloned stream enforce stream reading where i'm not a fan, will do a better implementation for the wanted use case.

PR #50, opened 2016-07-28
Added cURL formatter

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #19 
| Documentation   | 
| License         | MIT

#### What's in this PR?

This will add a cURL formatter

### DIFF
--- a/src/Formatter/CurlCommandFormatter.php
+++ b/src/Formatter/CurlCommandFormatter.php
@@ -33,7 +33,7 @@ class CurlCommandFormatter implements Formatter
 
         $body = $request->getBody()->__toString();
         if (!empty($body)) {
-            $command .= sprintf(' --data \'%s\'', $body);
+            $command .= sprintf(' --data \'%s\'', escapeshellarg($body));
         }
 
         return $command;
--- a/src/Formatter/CurlCommandFormatter.php
+++ b/src/Formatter/CurlCommandFormatter.php
@@ -3,7 +3,6 @@
 namespace Http\Message\Formatter;
 
 use Http\Message\Formatter;
-use Psr\Http\Message\MessageInterface;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 
@@ -21,20 +20,20 @@ class CurlCommandFormatter implements Formatter
     {
         $command = sprintf('curl \'%s\'', $request->getUri());
         if ($request->getProtocolVersion() === '1.0') {
-            $command.=' --http1.0';
+            $command .= ' --http1.0';
         } elseif ($request->getProtocolVersion() === '2.0') {
-            $command.=' --http2';
+            $command .= ' --http2';
         }
 
         $command .= ' --request '.$request->getMethod();
 
         foreach ($request->getHeaders() as $name => $values) {
-            $command .= sprintf(' -H \'%s: %s\'',$name, $request->getHeaderLine($name));
+            $command .= sprintf(' -H \'%s: %s\'', $name, $request->getHeaderLine($name));
         }
 
         $body = $request->getBody()->__toString();
         if (!empty($body)) {
-            $command .= sprintf(' --data \'%s\'',$body);
+            $command .= sprintf(' --data \'%s\'', $body);
         }
 
         return $command;
--- a/src/Formatter/CurlCommandFormatter.php
+++ b/src/Formatter/CurlCommandFormatter.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Http\Message\Formatter;
+
+use Http\Message\Formatter;
+use Psr\Http\Message\MessageInterface;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * A formatter that prints a cURL command for HTTP requests.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class CurlCommandFormatter implements Formatter
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function formatRequest(RequestInterface $request)
+    {
+        $command = sprintf('curl \'%s\'', $request->getUri());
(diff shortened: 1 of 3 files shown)


src/Formatter/CurlCommandFormatter.php
```diff
@@ -0,0 +1,49 @@
...
+
+        $command .= ' --request '.$request->getMethod();
+
+        foreach ($request->getHeaders() as $name => $values) {
+            $command .= sprintf(' -H \'%s: %s\'', $name, $request->getHeaderLine($name));
+        }
+
+        $body = $request->getBody()->__toString();
+        if (!empty($body)) {
+            $command .= sprintf(' --data \'%s\'', $body);
```
joelwurtz: Maybe we should espace the body (what give a json string with ' or " ?, need a test for this)

src/Formatter/CurlCommandFormatter.php
```diff
@@ -0,0 +1,49 @@
...
+            $command .= ' --http2';
+        }
+
+        $command .= ' --request '.$request->getMethod();
+
+        foreach ($request->getHeaders() as $name => $values) {
+            $command .= sprintf(' -H \'%s: %s\'', $name, $request->getHeaderLine($name));
+        }
+
+        $body = $request->getBody()->__toString();
```
joelwurtz: We should really cloned the stream before formatting IMO, i cannot imagine the number of issue and debugging nightmare for users not understanding how streams work.

Nyholm: Hm, tell me why I should *clone* it? Shouldn't I just rewind it just like the `FullHttpMessageFormatter`?

joelwurtz: Rewind is not possible everwhere, i.e. in the socket-client you cannot rewind a stream, but don't handle that here, that was just a reminder.

Nyholm: Okey, So what do you propose? 

$body = clone $request->getBody()?

joelwurtz: https://github.com/php-http/message/pull/49 :)

src/Formatter/CurlCommandFormatter.php
```diff
@@ -31,9 +31,13 @@ public function formatRequest(RequestInterface $request)
             $command .= sprintf(' -H \'%s: %s\'', $name, $request->getHeaderLine($name));
         }
 
-        $body = $request->getBody()->__toString();
-        if (!empty($body)) {
-            $command .= sprintf(' --data \'%s\'', escapeshellarg($body));
+        $body = $request->getBody();
+        if ($body->getSize() > 0) {
+            if (!$body->isSeekable()) {
+                throw new \RuntimeException('Cannot take data from a stream that is not seekable');
```
joelwurtz: Sending an exception here is horrible, i.e. some requests can randomly fail only on debugging (not on prod)  in a symfony env with this formatter in the debug toolbar. Can we set a specific exception like : CannotFormatException (or something else) that we can catch in class using this formatter to not show something ?

Nyholm: I was planning to remove this code when your request cloner is merged.

src/Formatter/CurlCommandFormatter.php
```diff
@@ -0,0 +1,53 @@
...
+        $command .= ' --request '.$request->getMethod();
+
+        foreach ($request->getHeaders() as $name => $values) {
+            $command .= sprintf(' -H %s', escapeshellarg($name.': '.$request->getHeaderLine($name)));
+        }
+
+        $body = $request->getBody();
+        if ($body->getSize() > 0) {
+            if (!$body->isSeekable()) {
+                throw new \RuntimeException('Cannot take data from a stream that is not seekable');
```
dbu: is this handled outside? or should we just return something in the text that says there would be the body but we can't read it?

Nyholm: See https://github.com/php-http/message/pull/50#discussion_r72948472

dbu:
did you look at https://github.com/namshi/cuzzle ? if so we should probably mention the authors of that in the class docblock.

is there anything in namshi that we are missing here?

Nyholm:
I did not. Haven't seen this package until now. They have somethings we dont.. I'll update the PR.

Nyholm:
I've update the PR with some fixes borrowed from Cuzzle. See the [diff](https://github.com/php-http/message/pull/50/commits/df3912932bda7644be47fd22359f61f4a5385388). 

It is not too much code. Should I mention the Cuzzle authors?

sagikazarmark:
> Should I mention the Cuzzle authors?

Just mention in the README in a credits section that it is inspired by cuzzle.

https://github.com/php-http/guzzle6-adapter#credits

dbu:
thanks a lot! created https://github.com/php-http/documentation/issues/136 about documenting this. we might also want to enable it in HttplugBundle

Nyholm:
Thank you for merging. 

👍

src/Formatter/CurlCommandFormatter.php
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        $command .= $this->getHeadersAsCommandOptions($request);
+
+        $body = $request->getBody();
+        if ($body->getSize() > 0) {
+            if (!$body->isSeekable()) {
+                return 'Cant format Request as cUrl command if body stream is not seekable.';
+            }
+            $command .= sprintf(' --data %s', escapeshellarg($body->__toString()));
```
dbu: i wonder if we should have some sort of sanity check to not put huge files in here? say if the body size > 5000 we truncate or just do something like [too large body] in the command? if we use this formatter eg in the symfony debug toolbar, it could break on large files

sagikazarmark:
Thanks a lot @Nyholm

PR #52, opened 2016-08-04
Add buffered stream

joelwurtz:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | mentioned in #46
| Documentation   | 
| License         | MIT

This is a replacement for the MemoryCloner, this allow to decorate a not seekable stream with a seekable one.

In order to do that it buffer the underlying stream into a php://temp resource by default (but can also be fully buffered into memory).

Once this is merged i will add a EnsureSeekableStreamPlugin, which decorates the request body and the response body with this one if the existing stream is not seekable, so we would be able to get the stream content (format) even for not seekable stream (only if user want it).

### DIFF
--- a/src/Stream/BufferedStream.php
+++ b/src/Stream/BufferedStream.php
@@ -1,0 +1,264 @@
+<?php
+
+namespace Http\Message\Stream;
+
+use Psr\Http\Message\StreamInterface;
+
+/**
+ * A buffered stream allow to buffer over an existing.
+ *
+ * You should use this decorator when you want to seek over a not seekable stream.
+ * This stream is however read-only.
+ */
+class BufferedStream implements StreamInterface
+{
+    /** @var resource The buffered resource used to seek previous data */
+    private $resource;
+
+    /** @var integer size of the stream if available */
+    private $size;
+
+    /** @var StreamInterface The underlying stream decorated by this class */
+    private $stream;
+
+    /** @var int How many bytes were written */
+    private $writed = 0;
+
+    public function __construct(StreamInterface $stream, $useFileBuffer = true, $memoryBuffer = 2097152)
+    {
+        $this->stream = $stream;
+        $this->size = $stream->getSize();
+
+        if ($useFileBuffer) {
+            $this->resource = fopen('php://temp/maxmemory:'.$memoryBuffer, 'rw+');
+        } else {
+            $this->resource = fopen('php://memory', 'rw+');
+        }
+
+        if (false === $this->resource) {
+            throw new \RuntimeException('Cannot create a resource over temp or memory implementation');
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function __toString()
+    {
+        try {
+            $this->rewind();
+
+            return $this->getContents();
+        } catch (\Throwable $throwable) {
+            return '';
+        } catch (\Exception $exception) {
+            return '';
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function close()
+    {
+        if (null === $this->resource) {
+            throw new \RuntimeException('Cannot close on a detached stream');
+        }
+
+        $this->stream->close();
+        fclose($this->resource);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function detach()
+    {
+        if (null === $this->resource) {
+            return null;
+        }
+
+        // Force reading the remaining data of the stream
+        $this->getContents();
+
+        $resource = $this->resource;
+        $this->stream = null;
+        $this->resource = null;
+
+        return $resource;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSize()
+    {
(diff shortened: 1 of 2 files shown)


src/Stream/BufferedStream.php
```diff
@@ -0,0 +1,264 @@
...
+    private $resource;
+
+    /** @var int size of the stream if available */
+    private $size;
+
+    /** @var StreamInterface The underlying stream decorated by this class */
+    private $stream;
+
+    /** @var int How many bytes were written */
+    private $writed = 0;
```
Nyholm: should be named to $written

src/Stream/BufferedStream.php
```diff
@@ -0,0 +1,264 @@
+<?php
+
+namespace Http\Message\Stream;
+
+use Psr\Http\Message\StreamInterface;
+
+/**
+ * A buffered stream allow to buffer over an existing.
```
dbu: lets make this summary: "Decorator to make any stream seakable."

src/Stream/BufferedStream.php
```diff
@@ -0,0 +1,264 @@
...
+ *
+ * You should use this decorator when you want to seek over a not seekable stream.
+ * This stream is however read-only.
```
dbu: we could mention how this decorator buffers the stream and warn that really large streams could lead to memory problems. (or is it writing to disk if the stream is too large for memory?)

src/Stream/BufferedStream.php
```diff
@@ -0,0 +1,264 @@
...
+    {
+        if (null === $this->resource) {
+            return;
+        }
+
+        // Force reading the remaining data of the stream
+        $this->getContents();
+
+        $resource = $this->resource;
+        $this->stream = null;
```
dbu: should we close the stream instead of just setting it to null? it is copied into $resource now

src/Stream/BufferedStream.php
Nyholm: Please correct me if Im wrong but I do not think it would ever lead to a memory problem. We use max 2MB memory as default (see constructor)

src/Stream/BufferedStream.php
```diff
@@ -0,0 +1,264 @@
...
+
+        return $resource;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSize()
+    {
+        if (null === $this->size && $this->stream->eof()) {
```
dbu: this->stream could be null when we detached previously

src/Stream/BufferedStream.php
```diff
@@ -0,0 +1,264 @@
...
+            if (null === $this->resource) {
```
dbu: should this check not be outside the other if? or do we return a size if we have one even if resource has since been set to null?

src/Stream/BufferedStream.php
dbu: maybe $this->eof()?

src/Stream/BufferedStream.php
dbu: right, looks like that. still, it would be nice to summarize how this works and what the implications are in the class doc.

src/Stream/BufferedStream.php
joelwurtz: Yes let's just return null if detached

src/Stream/BufferedStream.php
```diff
@@ -0,0 +1,270 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function __toString()
+    {
+        try {
+            $this->rewind();
+
+            return $this->getContents();
+        } catch (\Throwable $throwable) {
```
dbu: isn't this php 7 only? ah if it does not exist, php 5 will ignore this and use the catch below. maybe add a comment below that thats only for BC with php 5 and can be removed when going to php 7

joelwurtz: done

joelwurtz:
Is it ok to merge ?

dbu:
looks good to me. @Nyholm @sagikazarmark do you agree?

sagikazarmark:
AFAIK there are some PSR7 compliance test suits out there. Maybe we should use them to test PSR7 implementations. If they are not good enough, maybe we could write one? Apart from that, LGTM.

Nyholm:
👍

dbu:
more testing is always good, but lets not hold up merging. @joelwurtz if you have time to check out the idea from mark, that would be great to detect issues before we release.

joelwurtz:
I only found https://github.com/Maks3w/Psr7Assertions but this does not test stream

sagikazarmark:
I was once thinking about writting a PSR7 tester lib which accepts message factories to create messages/streams so that it can handle specific data and other options. The other possibility is to have an abstract method which returns a PSR7 object instance and that can be used since its immutable, although there would be some drawbacks like streams are noz immutable, etc.

PR #53, opened 2016-10-13
Add message, stream and URI factories for Slim Framework

tuupola:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | https://github.com/php-http/documentation/pull/159
| License         | MIT

#### What's in this PR?

This PR adds message, stream and URI factories for [Slim Framework](https://github.com/slimphp/Slim) 

#### Why?

Slim is reasonably popular PSR-7 based framework.

#### Example Usage

``` php
use Http\Client\Curl\Client;
use Http\Message\MessageFactory\SlimMessageFactory;
use Http\Message\StreamFactory\SlimStreamFactory;

$client = new Client(new SlimMessageFactory(), new SlimStreamFactory());
```

#### Checklist

- [x] Documentation pull request created (if not simply a bugfix)

#### To Do

- [ ] Should something be added to [puli.json](https://github.com/php-http/message/blob/master/puli.json)?
- [ ] Should these be named `Slim3StreamFactory` and `Slim3StreamFactory` instead?
- [ ] Should there be more tests? This was so easy I feel like I am missing something.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,9 @@ before_install:
 install:
     - travis_retry composer update ${COMPOSER_FLAGS} --prefer-dist --no-interaction
 
+before_script:
+    - if [[ "${TRAVIS_PHP_VERSION}" == "5.4" ]]; then composer remove slim/slim; fi
+
 script:
     - $TEST_COMMAND
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,13 +30,11 @@ matrix:
 
 before_install:
     - if [[ $COVERAGE != true ]]; then phpenv config-rm xdebug.ini || true; fi
+    - if [[ "${TRAVIS_PHP_VERSION}" == "5.4" ]]; then composer remove slim/slim --dev --no-update; fi
     - travis_retry composer self-update
 
 install:
     - travis_retry composer update ${COMPOSER_FLAGS} --prefer-dist --no-interaction
-
-before_script:
-    - if [[ "${TRAVIS_PHP_VERSION}" == "5.4" ]]; then composer remove slim/slim; fi
 
 script:
     - $TEST_COMMAND
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
     "suggest": {
         "zendframework/zend-diactoros": "Used with Diactoros Factories",
         "guzzlehttp/psr7": "Used with Guzzle PSR-7 Factories",
-        "slim/slim": "Used with Slim Factories",
+        "slim/slim": "Used with Slim Framework PSR-7 implementation",
         "ext-zlib": "Used with compressor/decompressor streams"
     },
     "autoload": {
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
         "php": ">=5.4",
         "psr/http-message": "^1.0",
         "php-http/message-factory": "^1.0.2",
-        "clue/stream-filter": "^1.3"
+        "clue/stream-filter": "^1.3",
+        "slim/slim": "^3.5"
     },
     "require-dev": {
         "zendframework/zend-diactoros": "^1.0",
@@ -23,7 +24,7 @@
         "phpspec/phpspec": "^2.4",
         "henrikbjorn/phpspec-code-coverage" : "^1.0",
         "coduo/phpspec-data-provider-extension": "^1.0",
-        "slim/slim": "^3.5"
+        "akeneo/phpspec-skip-example-extension": "^1.0"
     },
     "suggest": {
         "zendframework/zend-diactoros": "Used with Diactoros Factories",
--- a/composer.json
+++ b/composer.json
@@ -14,8 +14,7 @@
         "php": ">=5.4",
         "psr/http-message": "^1.0",
         "php-http/message-factory": "^1.0.2",
-        "clue/stream-filter": "^1.3",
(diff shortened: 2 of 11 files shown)


tuupola:
Build fails on PHP 5.4 because Slim requires PHP >=5.5.0.

sagikazarmark:
@tuupola thank you for the PR. Can you try adding https://github.com/akeneo/PhpSpecSkipExampleExtension and skip the tests on PHP 5.4?

tuupola:
That might not be enough since also `composer install` fails. I see what I can do with the Travis build matrix.

sagikazarmark:
@php-http/owners how about dropping PHP 5.4 support of this package too?

Nyholm:
Dropping 5.4 is a no-brainer in my opinion. My general approach is that you should keep old version PHP if there is no reson for bumping then. I believe @tuupola has an excellent reson here. 

We had a [discussion on twitter](https://twitter.com/xabbuh/status/781046823251218432) about supporting 5.4 because legacy project would not be able to update to HTTPlug without dropping support for 5.4. (Because of php-http/discovery).

xabbuh:
As this library is only a dependency during development I would not use it as an argument to bump the minimum version requirements. Instead I would remove the dev requirement and install the package on Travis CI only for PHP >= 5.5 jobs and skip the example otherwise as @sagikazarmark suggested.

sagikazarmark:
@xabbuh the problem is that this way development is much harder. I would do the other way around: remove the library in CI on PHP 5.4. That way you can't develop it if you have 5.4, but the lib remains compatible with it.

Nyholm:
Sure, Im fine with that.

xabbuh:
@sagikazarmark Sounds good too.

tuupola:
I already have it the way @xabbuh explained. However comment by @sagikazarmark makes sense. I will do it over and commit soonish.

.travis.yml
```diff
@@ -35,6 +35,9 @@ before_install:
 install:
     - travis_retry composer update ${COMPOSER_FLAGS} --prefer-dist --no-interaction
 
+before_script:
+    - if [[ "${TRAVIS_PHP_VERSION}" == "5.4" ]]; then composer remove slim/slim; fi
```
sagikazarmark: I would move this to before install AND add --no-update flag to composer. That way we only have to run compose once.

composer.json
```diff
@@ -14,19 +14,22 @@
         "php": ">=5.4",
         "psr/http-message": "^1.0",
         "php-http/message-factory": "^1.0.2",
-        "clue/stream-filter": "^1.3"
+        "clue/stream-filter": "^1.3",
+        "slim/slim": "^3.5"
```
sagikazarmark: This is only dev requirement, right?

tuupola: Should be yes. But this way it wont still work because. You cannot do the `composer install` and then `composer remove slim/slim` with PHP 5.4 anyway since it is the  `composer install` which  will fail. I try if `--ignore-platform-reqs` helps.

sagikazarmark: Not sure I understand you. Also, please see my previous comment about moving the removal to before_install instead of before_script with --no-update. --no-update causes to update composer.json, but does not actually call install.

composer.json
```diff
@@ -22,11 +22,14 @@
...
         "henrikbjorn/phpspec-code-coverage" : "^1.0",
-        "coduo/phpspec-data-provider-extension": "^1.0"
+        "coduo/phpspec-data-provider-extension": "^1.0",
+        "akeneo/phpspec-skip-example-extension": "^1.0",
+        "slim/slim": "^3.5"
     },
     "suggest": {
         "zendframework/zend-diactoros": "Used with Diactoros Factories",
         "guzzlehttp/psr7": "Used with Guzzle PSR-7 Factories",
+        "slim/slim": "Used with Slim Factories",
```
sagikazarmark: Diactoros and Guzzle PSR-7 are obviously meaning message factories while Slim is a framework. Can you please add some extra indication that it means message/PSR-7 factories? Just to avoid confusion.

tuupola:
Seems there is problem with streams with hhvm. Will install hhvm locally to figure it out.

composer.json
tuupola: Not sure about wording. Maybe `"Used with Slim Framework PSR-7 implementation"` or  `"Used with Slim Framework PSR-7 messages"`?

tuupola:
I would also like to add something to message factory tests. Namely I would like to test that factories set the headers and body correctly. Should this be a separate PR?

Nyholm:
Great job. Thank you!

sagikazarmark:
@Nyholm there were some TODOs which indeed would have been nice to have here.

Nyholm:
Sorry, I must have missed that. I saw that you approved it so I made a final review myself. 

What were the TODOs?

sagikazarmark:
See the TODO section: puli.json, possibly rename to Slim3

Nyholm:
Sorry for not bringing those up to discussion before merging. 

Yes, some entries should be added to Puli.json. The other two are fine IMHO

tuupola:
I am not familiar with Puli. How should I add the entries to puli.json? I quickly read through puli docs but did not become any smarter.

PR #59, opened 2016-10-19
Remove encoding writing stuff

ajgarlag:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    | no|
| BC breaks?      | no
| Deprecations?   | yes
| License         | MIT

#### What's in this PR?

This PR removes all stuff related with write filters in the `Http\Message\Encoding` namespace.

#### Why?

The abstract class `Http\Message\Encoding\FilteredStream` defines an abstract method named `getWriteFilter` that is not called anywhere and produces dead code not being tested. IMO this code must be removed or at least deprecated.

#### Example Usage

You can run the tests to see that this code is not being used nor tested.
``` 
$composer test
```

#### Checklist

#### To Do

- [x] Squash the commits before merge

### DIFF
--- a/src/Encoding/FilteredStream.php
+++ b/src/Encoding/FilteredStream.php
@@ -24,16 +24,22 @@ abstract class FilteredStream implements StreamInterface
 
     /**
      * @var resource
+     *
+     * @deprecated since version 1.5, will be removed in 2.0
      */
     protected $readFilter;
 
     /**
      * @var callable
+     *
+     * @deprecated since version 1.5, will be removed in 2.0
      */
     protected $writeFilterCallback;
 
     /**
      * @var resource
+     *
+     * @deprecated since version 1.5, will be removed in 2.0
      */
     protected $writeFilter;
 
@@ -47,12 +53,17 @@ abstract class FilteredStream implements StreamInterface
     /**
      * @param StreamInterface $stream
      * @param mixed|null      $readFilterOptions
-     * @param mixed|null      $writeFilterOptions
+     * @param mixed|null      $writeFilterOptions deprecated since 1.5, will be removed in 2.0
      */
     public function __construct(StreamInterface $stream, $readFilterOptions = null, $writeFilterOptions = null)
     {
-        $this->readFilterCallback = Filter\fun($this->getReadFilter(), $readFilterOptions);
-        $this->writeFilterCallback = Filter\fun($this->getWriteFilter(), $writeFilterOptions);
+        $this->readFilterCallback = Filter\fun($this->readFilter(), $readFilterOptions);
+        $this->writeFilterCallback = Filter\fun($this->writeFilter(), $writeFilterOptions);
+
+        if (null !== $writeFilterOptions) {
+            @trigger_error('The $writeFilterOptions argument is deprecated since version 1.5 and will be removed in 2.0.', E_USER_DEPRECATED);
+        }
+
         $this->stream = $stream;
     }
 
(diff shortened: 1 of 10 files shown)


sagikazarmark:
As far as I can see it's being called here: https://github.com/php-http/message/blob/master/src/Encoding/FilteredStream.php#L55

ajgarlag:
Yes, but the `writeFilterCallback` is not being called anywhere

sagikazarmark:
That's true, but technically the methods you removed are still called.

sagikazarmark:
Ping @joelwurtz

joelwurtz:
My intent was to propose a method to write into a filtered stream which is the reverse filter of the encoding one.

This can be useful on server side or library manipulating stream (like transfering data from a FilteredStream to another FilteredStream should result in the same data, which is not the case actually as it would only decode multiple times but never reencode)

However i nerver implemented the method as ATM there was no need for that, removing this is not a problem as like you said it's never used (and it's not a BC break). 

But there should be more to that, the filtered stream should only be readable and not writable (so isWritable should return false, and write method should throw an exception) and also it should not be seekable (but not the purpose of this PR, can be reported as a separate Issue or write as a separate PR)

sagikazarmark:
Well, the fact that we don't use it is enough for removal?

Also, IMO it's a BC break, no matter if it's used by US or not, it's a public API change.

dbu:
this is a massive BC break in our public API. if we decide we don't need this we should deprecate the methods and remove them in version 2 only. if they are useful as is, we should instead write some tests that make sure they work as intended.

joelwurtz:
Ah yes didn't see the public visibility, i think it's a mistake in the original code as user never need to get the encoding or decoding filter it's only an implementation detail, they should be deprecated then and removed in 2.0.

dbu:
so we can keep this PR open but mark it to be version 2 only, and do
another one to deprecate the methods? should we also trigger a
deprecation warning like symfony does?

ajgarlag:
@dbu I've already modified the PR to deprecate the methods.

src/Encoding/FilteredStream.php
```diff
@@ -47,12 +32,14 @@
...
      * @param mixed|null      $readFilterOptions
-     * @param mixed|null      $writeFilterOptions
+     * @param mixed|null      $writeFilterOptions deprecated since 1.5, will be removed in 2.0
      */
     public function __construct(StreamInterface $stream, $readFilterOptions = null, $writeFilterOptions = null)
     {
-        $this->readFilterCallback = Filter\fun($this->getReadFilter(), $readFilterOptions);
-        $this->writeFilterCallback = Filter\fun($this->getWriteFilter(), $writeFilterOptions);
+        $this->readFilterCallback = Filter\fun($this->readFilter(), $readFilterOptions);
+        if (null !== $writeFilterOptions) {
```
sagikazarmark: Can you please add some spacing before and after?

src/Encoding/FilteredStream.php
```diff
@@ -47,12 +32,14 @@
...
+            @trigger_error('The $writeFilterOptions argument is deprecated since version 1.5 and will be removed in 2.0. Use Http\Message\RequestMatcher\RequestMatcher instead.', E_USER_DEPRECATED);
```
sagikazarmark: You mean Http\Message\RequestMatcher? (interface).

sagikazarmark: Also, maybe use sprintf and ::class constant for code brevity?

src/Encoding/FilteredStream.php
```diff
@@ -23,21 +23,6 @@
...
 
     /**
-     * @var resource
-     */
-    protected $readFilter;
-
-    /**
-     * @var callable
-     */
-    protected $writeFilterCallback;
```
sagikazarmark: Since this is an abstract class: isn't this a BC break too?

Nyholm: Removing a non-private function in a non-final class is always a BC break.

ajgarlag: In that case, the `$readFilter` and `$writeFilter` properties should remain and should be marked as deprecated too.

ajgarlag:
@sagikazarmark ready for review

sagikazarmark:
LGTM

src/Encoding/FilteredStream.php
```diff
@@ -24,16 +24,22 @@
...
+     * @deprecated since version 1.5, will be removed in 2.0
      */
     protected $writeFilterCallback;
 
     /**
      * @var resource
+     *
+     * @deprecated since version 1.5, will be removed in 2.0
      */
     protected $writeFilter;
```
joelwurtz: I think this can be safely removed as it was always null and never used, even if it's BC i assume nobody use this (otherwise i really don't understand why xD)

@dbu @sagikazarmark ok to remove this even if it's somehow BC (never used, always null value in a protected scope so ...) ?

src/Encoding/FilteredStream.php
```diff
@@ -24,16 +24,22 @@
 
     /**
      * @var resource
+     *
+     * @deprecated since version 1.5, will be removed in 2.0
      */
     protected $readFilter;
```
joelwurtz: Can be removed as well as the writeFilter

src/Encoding/FilteredStream.php
dbu: i tend to agree, but lets maybe leave as is, there is stuff to remove when we go 2.0 anyways so we can play it safe. does not add cost to leave those in

PR #61, opened 2016-10-20
Fix encoding stream size

### LINKED ISSUE #60: Invalid value returned by GzipEncodeStream::getSize
| Q | A |
| --- | --- |
| Bug? | yes |
| New Feature? | no |
| Version | b86b453 |
#### Actual Behavior

The `GzipEncodeStream::getSize` returns the size of the inner stream content.
#### Expected Behavior

The `GzipEncodeStream::getSize` should return the size of the gzipped body.
#### Steps to Reproduce

``` php
<?php
require 'vendor/autoload.php';

$payload = 'This is a test stream';

$innerStream = new spec\Http\Message\Encoding\MemoryStream($payload);
$gzipEncodedStream = new Http\Message\Encoding\GzipEncodeStream($innerStream);

var_dump($gzipEncodedStream->getContents() === gzencode($payload)); //It is true
var_dump($gzipEncodedStream->getSize() === strlen(gzencode($payload))); //Should be true
```

ajgarlag:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| Related tickets | fixes #60 |
| License | MIT |
#### What's in this PR?

This PR includes the tests required to expose the problem described in #60 about the invalid `getSize` returned value by `FilteredStream` subclasses, and should include a fix.
#### Why?
#60 Invalid value returned by FilteredStream::getSize
#### Checklist
#### To Do
- [x] Discuss and find a solution to fix the tests
- [x] Squash commits

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## Fixed
+
+ - FilteredStream::getSize returns null because the contents size is unknown.
+
 ### Deprecated
 
  - FilteredStream::getReadFilter The read filter is internal and should never be used by consuming code.
--- a/spec/Encoding/ChunkStreamSpec.php
+++ b/spec/Encoding/ChunkStreamSpec.php
@@ -39,4 +39,12 @@ class ChunkStreamSpec extends ObjectBehavior
 
         $this->getContents()->shouldReturn("6\r\nThis i\r\n6\r\ns a st\r\n4\r\nream\r\n0\r\n\r\n");
     }
+
+    function it_does_not_know_the_content_size()
+    {
+        $stream = new MemoryStream('This is a stream');
+        $this->beConstructedWith($stream, 6);
+
+        $this->getSize()->shouldReturn(null);
+    }
 }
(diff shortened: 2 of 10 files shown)


ajgarlag:
A quick fix could be to implement a `getSize` method in `FilteredStream` to always return `null` (allowed by PSR-7 when the size is unknown) and rewrite the tests to expect that `null` value.

joelwurtz:
AFAIK, result size on the compression cannot be determined, but they may be headers for the decompression (not sure).

So :+1: for the null value as i don't see a way to do this unless we read up the whole stream ^^ which will make it non readable as we cannot seek a filtered stream (don't remember where is see this but using \php_user_filter make a stream not seekable)

dbu:
that looks at least not wrong to me. i am :+1: to merge this. for future reference, can we do better than this for some of the stream types?

joelwurtz:
LGTM,

Could you rebase with the last PR merged (to remove conflict in the changelog), then it's fine for me

sagikazarmark:
Sorry for the late response. Thanks.

PR #66, opened 2016-12-16
Fix cookie root path match for subdirectories. Refs #65

### LINKED ISSUE #65: Cookie for root path (/) matches only exact match.
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | v1.4.0

#### Actual Behavior

When having a Cookie with the path set to the domain root (e.g. `/`), and checking to see if another path matches (i.e. `/cookies`) it will always fail. (Except when checking an exact match, `/`).

#### Expected Behavior

I expect path `/cookies` to match a Cookie with path `/`.

#### Steps to Reproduce

```php
$cookie = new Cookie('foo', 'bar', null, 'example.com', '/');
var_dump($cookie->matchPath('/cookies')); // = false, but should be true
```

```php
// spec\Http\Message\CookieSpec
function it_matches_the_root_path()
{
    $this->beConstructedWith('name', 'value', null, null, '/');
    $this->matchPath('/')->shouldReturn(true);
    // we should add this line below
    $this->matchPath('/cookies')->shouldReturn(true);
}
```

#### Possible Solutions

A solution shouldn't be to hard to find. My guess is that adding a trailing slash isn't completely correct, but reading the linked [rfc section](http://tools.ietf.org/html/rfc6265#section-5.1.4) doesn't make things easier. 😕 

**current code. not a solution!:**
```php
/**
 * Checks whether this cookie is meant for this path.
 * @see http://tools.ietf.org/html/rfc6265#section-5.1.4
 */
public function matchPath($path)
{
    return $this->path === $path || (strpos($path, $this->path.'/') === 0);
}
```

Maff-:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no?
| Deprecations?   | no
| Related tickets | fixes #65
| License         | MIT

#### What's in this PR?

Fixes the problem where `(new Cookie('foo', 'bar', null, null, '/'))->matchPath('/cookie')` would return `false`.

#### Checklist

- [x] Updated CHANGELOG.md to describe bugfix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
+- Cookie::matchPath Cookie with root path (`/`) will not match sub path (e.g. `/cookie`).
+
 ### Deprecated
 
  - FilteredStream::getReadFilter The read filter is internal and should never be used by consuming code.
--- a/spec/CookieSpec.php
+++ b/spec/CookieSpec.php
@@ -189,6 +189,7 @@ class CookieSpec extends ObjectBehavior
         $this->beConstructedWith('name', 'value', null, null, '/path/to/somewhere');
 
         $this->matchPath('/path/to/somewhere')->shouldReturn(true);
+        $this->matchPath('/path/to/somewhere/else')->shouldReturn(true);
         $this->matchPath('/path/to/somewhereelse')->shouldReturn(false);
     }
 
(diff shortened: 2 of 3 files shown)


sagikazarmark:
Hi there,

Thanks for the PR.

Since the master branch already has some deprecations, which would require an 1.5 release, I cherry-picked your commit to an 1.4.x branch, you can see it here: https://github.com/php-http/message/commit/5fd6c761b994b16afaae4a236abb650e5a068b5a

sagikazarmark:
And tagged 1.4.1

Maff-:
Thanks for the quick action!

PR #68, opened 2016-12-22
Optional cookie validation

### LINKED ISSUE #67: Optional cookie validation
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 69e47233956fdbba8c32a18ecd35ee454626e6ea

#### Actual Behavior

I'm writing a proxy script where I have to parse the set-cookie headers sent back by the upstream servers. Sometimes the upstream server (out of my control) send a cookie with invalid characters in name and/or value and when I try to parse it the library throws an \InvalidArgumentException

#### Expected Behavior

I'd like to make this validation optional

#### Steps to Reproduce

```php
use Http\Discovery\HttpClientDiscovery;
use Http\Message\CookieJar;
use Http\Client\Common\PluginClient;
use Http\Client\Common\Plugin\CookiePlugin;
use Http\Client\Curl\Client;
use Http\Message\MessageFactory\DiactorosMessageFactory;
use Http\Message\StreamFactory\DiactorosStreamFactory;

$messageFactory = new DiactorosMessageFactory();

$pluginClient = new PluginClient(
    new Client($messageFactory, new DiactorosStreamFactory()),
    [new CookiePlugin(new CookieJar())]
);

$response = $pluginClient->sendRequest(
    $messageFactory->createRequest('GET', 'http://www.micromedexsolutions.com/home/dispatch')
);
```

ajgarlag:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #67
| License         | MIT

#### What's in this PR?

This PR adds optional cookie validation

#### Example Usage

TBD

#### Checklist

- [x] Documentation pull request created (if not simply a bugfix)

#### To Do

- [x] Discuss how to avoid current cookie constructor validation: additional flag parameter or using `ReflectionClass::newInstanceWithoutConstructor`

### DIFF
--- a/spec/CookieSpec.php
+++ b/spec/CookieSpec.php
@@ -244,24 +244,18 @@ class CookieSpec extends ObjectBehavior
         $this->isValid()->shouldReturn(true);
     }
 
-    /**
-     * @dataProvider invalidCharacterExamples
-     */
-    function it_can_be_constructed_without_name_validation($name, $invalid)
+    function it_can_be_constructed_without_name_validation()
     {
-        $this->beConstructedThrough('createWithoutValidation', [$name]);
+        $this->beConstructedThrough('createWithoutValidation', ["\x20"]);
 
-        $this->isValid()->shouldReturn(!$invalid);
+        $this->isValid()->shouldReturn(false);
     }
 
-    /**
-     * @dataProvider invalidCharacterExamples
-     */
-    function it_can_be_constructed_without_value_validation($value, $invalid)
+    function it_can_be_constructed_without_value_validation()
     {
-        $this->beConstructedThrough('createWithoutValidation', ['name', $value]);
+        $this->beConstructedThrough('createWithoutValidation', ['name', "\x20"]);
 
-        $this->isValid()->shouldReturn(!$invalid);
+        $this->isValid()->shouldReturn(false);
     }
 
     function it_can_be_constructed_without_max_age_validation()
--- a/spec/CookieSpec.php
+++ b/spec/CookieSpec.php
@@ -239,6 +239,38 @@ class CookieSpec extends ObjectBehavior
         $this->match($notMatches)->shouldReturn(false);
     }
 
+    function it_validates_itself()
+    {
+        $this->isValid()->shouldReturn(true);
+    }
+
+    /**
+     * @dataProvider invalidCharacterExamples
+     */
+    function it_can_be_constructed_without_name_validation($name, $invalid)
+    {
+        $this->beConstructedThrough('createWithoutValidation', [$name]);
+
+        $this->isValid()->shouldReturn(!$invalid);
+    }
+
+    /**
+     * @dataProvider invalidCharacterExamples
+     */
+    function it_can_be_constructed_without_value_validation($value, $invalid)
+    {
+        $this->beConstructedThrough('createWithoutValidation', ['name', $value]);
+
+        $this->isValid()->shouldReturn(!$invalid);
+    }
+
+    function it_can_be_constructed_without_max_age_validation()
+    {
+        $this->beConstructedThrough('createWithoutValidation', ['name', 'value', '-1']);
+
+        $this->isValid()->shouldReturn(false);
+    }
+
     /**
      * Provides examples for invalid characers in names and values.
      *
--- a/src/Cookie.php
+++ b/src/Cookie.php
@@ -54,6 +54,13 @@ final class Cookie
     private $expires;
 
     /**
+     * Validation state.
+     *
+     * @var bool
+     */
+    private $valid;
+
+    /**
      * @param string         $name
      * @param string|null    $value
      * @param int            $maxAge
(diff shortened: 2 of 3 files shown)


ajgarlag:
I've implemented it with an additional __construct parameter.

Maybe, we can rename `createWithoutValidation` to `create` and deprecate `__construct` making it private in 2.0

What do you think?

src/Cookie.php
```diff
@@ -73,11 +81,16 @@ public function __construct(
         $path = null,
         $secure = false,
         $httpOnly = false,
-        \DateTime $expires = null
+        \DateTime $expires = null,
+        $requireValidation = true
```
sagikazarmark: I don't really like the idea of having optional validation. Have it or not IMO. So I would just deprecate the exception and bypass the validation manually.

src/Cookie.php
```diff
@@ -89,6 +102,19 @@ public function __construct(
...
+        $name,
+        $value = null,
+        $maxAge = null,
+        $domain = null,
+        $path = null,
+        $secure = false,
+        $httpOnly = false,
+        \DateTime $expires = null
+    ) {
+        return new self($name, $value, $maxAge, $domain, $path, $secure, $httpOnly, $expires, false);
```
sagikazarmark: I would do some hacks here: pass null or the valid empty value to bypass the validation, then set the property directly. Since we are in the same class, we can set private properties.

src/Cookie.php
```diff
@@ -381,6 +407,27 @@ public function match(Cookie $cookie)
     }
 
     /**
+     * Validates cookie attributes.
+     *
+     * @return bool
+     */
+    public function isValid()
+    {
+        if (null === $this->valid) {
```
sagikazarmark: Hm, not sure about this. This value should be cleared at least when the value or the max age is "modified", but even then I am not sure the performance gain worth risking inconsistency.

ajgarlag: You are right, I'll remove this.

src/Cookie.php
```diff
@@ -63,7 +63,7 @@
      * @param bool           $httpOnly
      * @param \DateTime|null $expires  Expires attribute is HTTP 1.0 only and should be avoided.
      *
-     * @throws \InvalidArgumentException If name, value or max age is not valid.
+     * @throws \InvalidArgumentException If name, value or max age is not valid. Attributes validation during instantiation is deprecated since 1.5 and will be removed in 2.0.
```
sagikazarmark: I think it should be "Attribute validation"

src/Cookie.php
```diff
@@ -89,6 +94,24 @@ public function __construct(
         $this->httpOnly = (bool) $httpOnly;
     }
 
+    public static function createWithoutValidation(
```
sagikazarmark: Can you please add phpdoc?

sagikazarmark: Should this be deprecated as well? That would be weird, but still...

spec/CookieSpec.php
```diff
@@ -239,6 +239,38 @@ function it_matches_other_cookies()
         $this->match($notMatches)->shouldReturn(false);
     }
 
+    function it_validates_itself()
+    {
+        $this->isValid()->shouldReturn(true);
+    }
+
+    /**
+     * @dataProvider invalidCharacterExamples
```
sagikazarmark: Not sure if we want to test the fact that validation actually works correctly here. Although I can't really see any tests for that. Nevertheless I think we only need to test two cases here: validation passes/validation fails. For that I am not sure if we need data provider. In fact using data providers in specification tests is kind of abusing BDD, these kind of tests should be done in separate unit tests.

sagikazarmark:
@php-http/owners can you add some input here?

sagikazarmark:
> I do not think we should disable validation in the constructor though

My only concern is that the current way is a little bit hacky. Also, maybe we could make this a configuration in CookieJar (whether to check cookie validation or not)

sagikazarmark:
This looks good to me. Is there anything else @Nyholm ?

@ajgarlag can you please rebase your branch?

Nyholm:
Im happy with the current state of this PR. Thanks!

ajgarlag:
Rebased. Sorry for the delay.

sagikazarmark:
Does this require a follow up PR in the common-client repo to solve the original issue? Any last words @php-http/owners ?

ajgarlag:
I'll add a `$requireValidCookies` flag (`true` by default) to `CookieJar` and then, validate each cookie when is added to `CookieJar`. Then, in the common-client repo, I will always use `createWithoutValidation` to create the cookie object, although, I think this could be a BC in common-client.

Nyholm:
> I think this could be a BC in common-client

It **has** to be =)

I do not think the CookieJar should care about if the cookies are valid or not. The Jar is just a storage, it should be pretty stupid. 

I agree with you that the CookiePlugin should always use `createWithoutValidation`. Depending on configuration of the **CookiePlugin**, we validate the cookies before we put them in the jar. That is BC. 

👍  for merging this PR now.

ajgarlag:
> I agree with you that the CookiePlugin should always use createWithoutValidation.

As this the method that should be used **always** by CookiePlugin, I proposed:
> Maybe, we can rename createWithoutValidation to create and deprecate __construct making it private in 2.0

Nyholm:
I know. But making the constructor private would be a BC break. And we are doing it just because "*one class in a other library stopped using the constructor*". That is wrong. Also if we created (though unlikely) a `AlwaysValidCookiePlugin`, that class would use the constructor.

Nyholm:
Good to merge?

sagikazarmark:
Sorry for the long wait, thanks for the contribution @ajgarlag

PR #70, opened 2017-01-04
Add check for empty string to stream factories

### LINKED ISSUE #69: Check if body is empty string in stream factories
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes

#### Actual Behavior

Currently the stream factories check whether the body is null, and only in that case they don't write to the stream.

#### Expected Behavior

Empty strings are not written to a stream.

#### Possible Solutions

Writting an empty string to the stream doesn't make any sense, so we should check that too.

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #69
| Documentation   | does not apply
| License         | MIT

#### What's in this PR?

Stream factories now check empty strings as well, not just  null values.

### DIFF
--- a/src/StreamFactory/DiactorosStreamFactory.php
+++ b/src/StreamFactory/DiactorosStreamFactory.php
@@ -24,7 +24,7 @@ final class DiactorosStreamFactory implements StreamFactory
             } else {
                 $stream = new Stream('php://memory', 'rw');
 
-                if (null !== $body) {
+                if (null !== $body || '' !== $body) {
                     $stream->write((string) $body);
                 }
 
--- a/src/StreamFactory/DiactorosStreamFactory.php
+++ b/src/StreamFactory/DiactorosStreamFactory.php
@@ -24,9 +24,11 @@ final class DiactorosStreamFactory implements StreamFactory
             } else {
                 $stream = new Stream('php://memory', 'rw');
 
-                if (null !== $body || '' !== $body) {
-                    $stream->write((string) $body);
+                if (null === $body || '' === $body) {
+                    return $stream;
                 }
+
+                $stream->write((string) $body);
 
                 $body = $stream;
             }
--- a/src/StreamFactory/SlimStreamFactory.php
+++ b/src/StreamFactory/SlimStreamFactory.php
@@ -28,9 +28,11 @@ final class SlimStreamFactory implements StreamFactory
             $resource = fopen('php://memory', 'r+');
             $stream = new Stream($resource);
 
-            if (null !== $body || '' !== $body) {
-                $stream->write((string) $body);
+            if (null === $body || '' === $body) {
+                return $stream;
             }
+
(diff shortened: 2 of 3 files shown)


src/StreamFactory/DiactorosStreamFactory.php
```diff
@@ -24,7 +24,7 @@ public function createStream($body = null)
             } else {
                 $stream = new Stream('php://memory', 'rw');
 
-                if (null !== $body) {
+                if (null !== $body || '' !== $body) {
```
Nyholm: Why logical **or**? We should write to the stream if the body has content. 
```php
if (null !== $body && '' !== $body) {
```

src/StreamFactory/SlimStreamFactory.php
```diff
@@ -28,7 +28,7 @@ public function createStream($body = null)
             $resource = fopen('php://memory', 'r+');
             $stream = new Stream($resource);
 
-            if (null !== $body) {
+            if (null !== $body || '' !== $body) {
```
Nyholm: Same here. Use `&&`

src/StreamFactory/DiactorosStreamFactory.php
dbu: we could also do `if (null === $body || '' === $body) {return new Stream('php://memory', 'rw');}` as the rest of the code is irrelevant if the body is empty

joelwurtz: Also this doesn't work if body is an object and casting the object to string return an empty value. We should do the following : 

```
$content = (string) $body;

if (strlen($content) > 0) {
    $stream->write($content);
}

return $stream
```

sagikazarmark: @dbu indeed, an empty stream does not have to be rewinded, or does it?

@joelwurtz According to the interface the allowed types are `string|resource|StreamInterface|null`, so an object would not be valid in the first place, you would have to cast it to string first.

dbu:
> According to the interface the allowed types are string|resource|StreamInterface|null, so an object would not be valid in the first place, you would have to cast it to string first.

yes, but there is no type check to avoid objects. what exactly do we fix with this? is writing '' to a stream a bug? or is it just to be strict?

sagikazarmark:
Well, I guess you can technically pass objects there. There is a string casting before writting.

> is writing '' to a stream a bug? or is it just to be strict?

I don't think it's a bug. But we spare an unnecessary write (and now a rewind) operation. Not sure if adding an extra cast just to check if the string value of something is empty worth it.

dbu:
if objects are not supposed to be passed, i think we are good here. if you incorrectly pass an object, the behaviour does not change compared to previous, and should still not fail.

PR #71, opened 2017-01-16
Make sure we do not rewind a non-seekable stream

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | Relates to #72
| Documentation   | 
| License         | MIT

#### What's in this PR?

If we try to create a stream from a non-seekable resource we should not throw an exception.


joelwurtz:
I'm not sure we really want to rewind a stream, IMO this logic should be handled by the user providing the stream.

Nyholm:
I was unsure about this as well. That is why I created #72.

dbu:
the fix in itself is save, and good. lets discuss the general question of rewinding in #72

Nyholm:
Thank you for merging

PR #76, opened 2017-06-12
Implements a cookie-date parsing utility

aaa2000:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | mentioned in https://github.com/php-http/client-common/pull/46
| Documentation   | No
| License         | MIT

#### What's in this PR?

Implementation of a cookie-date parsing algorithm mentionned in https://github.com/php-http/client-common/pull/46. I preferred to use a simpler method from https://github.com/symfony/symfony/blob/master/src/Symfony/Component/BrowserKit/Cookie.php#L199.

#### Example Usage

``` php
$datetime = CookieUtil::parseDate('Friday, 31 Jul 20 08:49:37 GMT');
```

#### Checklist

- [ ] Documentation pull request created (if not simply a bugfix)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## Unreleased
 
+### Added
+
+- CookieUtil::parseDate to create a date from cookie date string
+
 ## 1.5.0 - 2017-02-14
 
 ### Added
--- a/src/CookieUtil.php
+++ b/src/CookieUtil.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace Http\Message;
+
+use Http\Message\Exception\UnexpectedValueException;
+
+final class CookieUtil
+{
+    /**
+     * Handles dates as defined by RFC 2616 section 3.3.1, and also some other
+     * non-standard, but common formats.
+     *
+     * @var array
+     */
+    private static $dateFormats = [
+        'D, d M y H:i:s T',
+        'D, d M Y H:i:s T',
+        'D, d-M-y H:i:s T',
+        'D, d-M-Y H:i:s T',
+        'D, d-m-y H:i:s T',
+        'D, d-m-Y H:i:s T',
+        'D M j G:i:s Y',
+        'D M d H:i:s Y T',
+    ];
+
+    /**
+     * @see https://github.com/symfony/symfony/blob/master/src/Symfony/Component/BrowserKit/Cookie.php
+     *
+     * @param string $dateValue
+     *
+     * @return \DateTime
+     *
+     * @throws UnexpectedValueException if we cannot parse the cookie date string.
+     */
+    public static function parseDate($dateValue)
+    {
+        foreach (self::$dateFormats as $dateFormat) {
+            if (false !== $date = \DateTime::createFromFormat($dateFormat, $dateValue, new \DateTimeZone('GMT'))) {
+                return $date;
+            }
+        }
+
+        // attempt a fallback for unusual formatting
+        if (false !== $date = date_create($dateValue, new \DateTimeZone('GMT'))) {
+            return $date;
+        }
+
+        throw new UnexpectedValueException(sprintf(
+            'Unparseable cookie date string "%s"',
+            $dateValue
+        ));
+    }
+}
--- a/spec/CookieUtilSpec.php
+++ b/spec/CookieUtilSpec.php
@@ -1,0 +1,76 @@
+<?php
+
+namespace spec\Http\Message;
+
+use Http\Message\Exception\UnexpectedValueException;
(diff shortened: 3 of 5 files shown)


src/CookieUtil.php
```diff
@@ -0,0 +1,40 @@
...
+final class CookieUtil
+{
+    /**
+     * Handles dates as defined by RFC 2616 section 3.3.1, and also some other
+     * non-standard, but common formats.
+     *
+     * @var array
+     */
+    private static $dateFormats = array(
+        'D, d M y H:i:s T',
```
aaa2000: I add the format "Tuesday, 31 Mar 99 07:42:12 GMT" at the line 14 mentionned in https://github.com/php-http/client-common/pull/46

src/CookieUtil.php
```diff
@@ -0,0 +1,40 @@
...
+        'D, d-m-Y H:i:s T',
+        'D M j G:i:s Y',
+        'D M d H:i:s Y T',
+    );
+
+    /**
+     * @see https://github.com/symfony/symfony/blob/master/src/Symfony/Component/BrowserKit/Cookie.php
+     *
+     * @param string $dateValue
+     * @return \DateTime|false
```
aaa2000: Should I rather throw an exception ?

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 
 ## Unreleased
 
+###
```
dbu: please use the title "Added" here

aaa2000: Fix

aaa2000:
@dbu if we use only date_create, the tests will fail for this format `Fri, 31-07-20 08:49:37 GMT`.

aaa2000:
But this format is not conform to the rfc https://tools.ietf.org/html/rfc6265#section-5.1.1

`
month = "Jan" | "Feb" | "Mar" | "Apr" | "May" | "Jun" | "Jul" | "Aug" | "Sep" | "Oct" | "Nov" | "Dec"
`

it has been added to manage unusual cookie dates, see commit https://github.com/symfony/browser-kit/commit/47f146d9406d695ed1ca8256b41542e367442d2c

aaa2000:
Rebase done

I added more tests from library mentioned in php-http/client-common#46 https://github.com/salesforce/tough-cookie/blob/master/test/date_test.js except `01 Jan 1600 00:00:00 GMT` which should not pass because the date before 1601.

Use only the function `date_create` all tests pass except `Fri, 31-07-20 08:49:37 GMT`. Format that does not appear to comply with RFC.

Maybe, `CookieUtil::parseDate` is not really useful.

aaa2000:
tests failed on PHP 5.4 and HHVM

aaa2000:
I will fix later


http://php.net/manual/en/class.datetime.php#datetime.changelog

```
5.4.24 	The COOKIE constant was changed to reflect RFC 1036 using a four digit year rather than a two digit year (RFC 850) as prior versions. 
```

dbu:
as hhvm works on master, did we just discover a difference between hhvm and php here?

the php 5.4 build fails on the same things, formatting the date: `expected "Friday, 31-Jul-2020 08:49...", but got "Friday, 31-Jul-2020 08:49...".` - unfortunately its cut off before any differences show up. 

it seems to me that `format(\DateTime::COOKIE)` uses different formats in different versions. as we don't want to test the date formatting, maybe just use an explicit formatting string so you can compare with a well known outcome?

src/CookieUtil.php
```diff
@@ -0,0 +1,41 @@
...
+        'D M j G:i:s Y',
+        'D M d H:i:s Y T',
+    ];
+
+    /**
+     * @see https://github.com/symfony/symfony/blob/master/src/Symfony/Component/BrowserKit/Cookie.php
+     *
+     * @param string $dateValue
+     *
+     * @return \DateTime|false
```
dbu: i would prefer throwing an exception when the date can't be parsed

aaa2000:
Yes it is `format(\DateTime::COOKIE)` https://3v4l.org/VpvRp, as you said, I will use use an explicit formatting string. I would do it tonight.

src/CookieUtil.php
aaa2000: a php exception like `UnexpectedValueException` or ``Http\Message\Exception\UnexpectedValueException` ?

aaa2000:
throws an exception when the date can't be parsed and commits squashed

dbu:
thanks a lot @aaa2000 !

Issue #78, opened 2017-06-21
False positive trigger of deprecation notice in FilteredStream


ste93cry:
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 1.4.1

#### Actual Behavior

When creating my own custom stream to format data into base64 format (are you interested in merging it into core?) I extended the `FilteredStream` class. The PHP built-in `convert.base64-encode` stream filter seems to not accept `null` as value of the `$readFilterOptions` and `$writeFilterOptions` arguments. I had to pass an empty array to work around the following error:

>Unable to access built-in filter: stream_filter_append(): unable to create or locate filter "convert.base64-encode" 

This fixes the problem, but I then get a deprecation notice saying this:

>The $writeFilterOptions argument is deprecated since version 1.5 and will be removed in 2.0.

#### Expected Behavior

As it seems that not all filters accept `null` as option of the `stream_filter_append` function, the deprecation notice should not be triggered when passing an empty array

#### Steps to Reproduce

Just create a class that extends the `FilteredStream` class and uses the `convert.base64-encode` stream filter without overriding the constructor. It will throw an exception when using it, and if constructor is overriden to fix the problem a deprecation notice will be triggered

#### Possible Solutions

The only solution I can think of is to check not only that the `$writeFilterOptions` argument is not `null` but also that it's a non-empty array to trigger the deprecation notice. Something like this:

```diff
diff --git a/src/Encoding/FilteredStream.php b/src/Encoding/FilteredStream.php
index a32554b..a516f77 100644
--- a/src/Encoding/FilteredStream.php
+++ b/src/Encoding/FilteredStream.php
@@ -60,7 +60,7 @@ abstract class FilteredStream implements StreamInterface
         $this->readFilterCallback = Filter\fun($this->readFilter(), $readFilterOptions);
         $this->writeFilterCallback = Filter\fun($this->writeFilter(), $writeFilterOptions);
 
-        if (null !== $writeFilterOptions) {
+        if (null !== $writeFilterOptions || (is_array($writeFilterOptions) && !empty($writeFilterOptions))) {
             @trigger_error('The $writeFilterOptions argument is deprecated since version 1.5 and will be removed in 2.0.', E_USER_DEPRECATED);
         }
```

sagikazarmark:
TBH I don't even remember why this was deprecated in the first place.

[...]

PR #79, opened 2017-06-23
Prepare for 1.6

Nyholm:


dbu:
would love if we could solve #78 before tagging 1.6 - lets give it another day or so but then proceed with the release to not block progress :-)

Nyholm:
Currently waiting for https://github.com/clue/php-stream-filter/pull/15 to be merged and released.

Nyholm:
Or should we wait for clue/php-stream-filter? Would we update our composer.json? `clue/php-stream-filter` is only going to release a patch. 

I think we should release 1.6 now.

PR #80, opened 2017-06-23
Added provide for php-http/message-factory-implementation

### LINKED ISSUE #75: Missing provides in composer.json file
| Q            | A
| ------------ | ---
| Bug?         | yes and no
| New Feature? | yes and no
| Version      | Latest

#### Actual Behavior
This package doesn't say it provides message factories (implementing `php-http/message-factory-implementation`). So if a package requires this implementation, it will not be installable, even though there are messages factories in this repository.

#### Expected Behavior
Stating that this package implements `php-http/message-factory-implementation`

#### Possible Solutions
Stating that this package implements `php-http/message-factory-implementation`

Even though it is also not technically correct as if one of these 3 factories can't be accessed (because their main vendor is not installed). Another solution would be to decouple these implementations into their own repositories (and stating then that it can implement said interface), or putting these in their client package (e.g for guzzle, putting its classes into the guzzle5 and guzzle6 adapters)

Nyholm:
This will fix #75

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,9 @@
         "php-http/message-factory": "^1.0.2",
         "clue/stream-filter": "^1.3"
     },
+    "provide": {
+        "php-http/message-factory-implementation": "1.0"
+    }
     "require-dev": {
         "zendframework/zend-diactoros": "^1.0",
         "guzzlehttp/psr7": "^1.0",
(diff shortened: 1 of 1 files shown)


sagikazarmark:
AFAIK we decided not to do this because these interfaces are somewhat internal and the most important implementations are in message. But I am not specifically against it, so LGTM.

PR #81, opened 2017-06-23
Prevent false positive triggering of deprecated notice in FilteredStream

ste93cry:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #78
| License         | MIT

#### What's in this PR?

In this PR the `$writeFilterOptions` argument of the `FilteredStream::__construct` method is normalized because some built-in PHP stream filters doesn't accept `null` as options. As a consequence of this, the deprecation message was triggered in some cases when it should not had to, and this has been addressed too. I'm not sure about the entry I added in the CHANGELOG, as I'm not a native English speaker maybe you can find a better phrase... I don't know if unit tests are needed for this PR as currently there is nothing hat tests the throwing of the error or the call to the `Filter\fun` function with the correct arguments. Ast a last thing, I can't test what options the `bzip2.compress` and `bzip2.decompress` accept, so the PR is not ready yet until someone finds out if `null` is a valid value or not.

#### Checklist

### DIFF
--- a/src/Encoding/FilteredStream.php
+++ b/src/Encoding/FilteredStream.php
@@ -57,10 +57,22 @@ abstract class FilteredStream implements StreamInterface
      */
     public function __construct(StreamInterface $stream, $readFilterOptions = null, $writeFilterOptions = null)
     {
+        $triggerDeprecationNotice = null !== $writeFilterOptions;
+
+        if (null === $readFilterOptions && in_array($this->readFilter(), ['convert.base64-encode', 'convert.base64-decode', 'convert.quoted-printable-encode', 'zlib.deflate', 'zlib.inflate'/*, 'bzip2.compress', 'bzip2.decompress'*/], true)) {
+            $readFilterOptions = [];
+        }
+
+        if (null === $writeFilterOptions && in_array($this->writeFilter(), ['convert.base64-encode', 'convert.base64-decode', 'convert.quoted-printable-encode', 'zlib.deflate', 'zlib.inflate'/*, 'bzip2.compress', 'bzip2.decompress'*/], true)) {
+            $writeFilterOptions = [];
+
+            $triggerDeprecationNotice = false;
+        }
+
         $this->readFilterCallback = Filter\fun($this->readFilter(), $readFilterOptions);
         $this->writeFilterCallback = Filter\fun($this->writeFilter(), $writeFilterOptions);
 
-        if (null !== $writeFilterOptions) {
+        if ($triggerDeprecationNotice) {
             @trigger_error('The $writeFilterOptions argument is deprecated since version 1.5 and will be removed in 2.0.', E_USER_DEPRECATED);
         }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
 
 - CookieUtil::parseDate to create a date from cookie date string
 
+### Fixed
+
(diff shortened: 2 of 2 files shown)


src/Encoding/FilteredStream.php
```diff
@@ -57,10 +57,22 @@
      */
     public function __construct(StreamInterface $stream, $readFilterOptions = null, $writeFilterOptions = null)
     {
+        $triggerDeprecationNotice = null !== $writeFilterOptions;
+
+        if (null === $readFilterOptions && in_array($this->readFilter(), ['convert.base64-encode', 'convert.base64-decode', 'convert.quoted-printable-encode', 'zlib.deflate', 'zlib.inflate'/*, 'bzip2.compress', 'bzip2.decompress'*/], true)) {
```
Nyholm: Instead of listing functions like this, we could do:

```php
if (empty($readFilterOptions)) {
  $this->readFilterCallback = Filter\fun($this->readFilter());
} else {  
}
```

src/Encoding/FilteredStream.php
```diff
@@ -57,10 +57,22 @@
...
+            $writeFilterOptions = [];
+
+            $triggerDeprecationNotice = false;
+        }
+
         $this->readFilterCallback = Filter\fun($this->readFilter(), $readFilterOptions);
         $this->writeFilterCallback = Filter\fun($this->writeFilter(), $writeFilterOptions);
 
-        if (null !== $writeFilterOptions) {
+        if ($triggerDeprecationNotice) {
```
Nyholm: We should always trigger a deprecation notice if the `$writeFilterOptions` is used. The `$this->writeFilterCallback` is deprecated because it is never used.

dbu:
@ste93cry when you address the feedback from tobias, please also rebase on master - there is a conflict with the CHANGELOG file that will prevent automatic merging.

src/Encoding/FilteredStream.php
ste93cry: Unfortunately the code as you wrote it won't work because `Filter\fun` by default pass `null` as value of the `$params` argument of the `stream_filter_append` function. I proposed a solution in #78 that I tested with the `base64_encode` filter (which is the reason I found the problem and opened the issue), but @dbu told me to explicitly check for the filter names. I don't know if there is a third solution, but at the moment the only two that I tested and works are these

src/Encoding/FilteredStream.php
ste93cry: This with the `if` statements that I added above does exactly this, but I can't trigger a deprecation notice when the value of `$writeFilterOptions` is the default value for its filter

Nyholm:
Thank you @ste93cry for your responses. 

What I do not like about this PR is that we list a bunch of filters (and it is missing tests). This is a "workaround" and not a "fix". 

I have [submitted some PRs](https://github.com/clue/php-stream-filter/pulls?q=is%3Apr+author%3ANyholm) to the clue/php-stream-filter in order to actually fix this problem. 
I hope you are okey with closing this PR update our dependencies to the next version of that library?

Also, unrelated bug: https://github.com/clue/php-stream-filter/pull/17

ste93cry:
>I hope you are okey with closing this PR update our dependencies to the next version of that library?

I agree that fixing this upstream is a better solution that this PR, so it's fine. Just for reference in case someone lands here in the future, the ticket related to this problem is clue/php-stream-filter#15

PR #86, opened 2017-11-23
Require php-stream-filter 1.4

### LINKED ISSUE #78: False positive trigger of deprecation notice in FilteredStream

Nyholm:
This will avoid some bugs in 1.3. This will fix #78

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "php": ">=5.4",
         "psr/http-message": "^1.0",
         "php-http/message-factory": "^1.0.2",
-        "clue/stream-filter": "^1.3"
+        "clue/stream-filter": "^1.4"
     },
     "provide": {
         "php-http/message-factory-implementation": "1.0"


dbu:
i just merged the styleci fix

PR #94, opened 2018-03-07
Bug/curl fromater fails

### LINKED ISSUE #93: CurlFormatter fails on large boddies
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | Master

#### Actual Behavior

When posting super large bodies you will get an error because escshellargs does not allow more than X chars. https://github.com/php-http/message/blob/v1.4.1/src/Formatter/CurlCommandFormatter.php#L42

We should make sure to catch/suppress that error and write a better error message in the curl string.

llaakkkk:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no|yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #93 
| License         | MIT

#### What's in this PR?

Added suppress to CurlFormatter for big file error.

### DIFF
--- a/src/Formatter/CurlCommandFormatter.php
+++ b/src/Formatter/CurlCommandFormatter.php
@@ -3,12 +3,13 @@
 namespace Http\Message\Formatter;
 
 use Http\Message\Formatter;
+use PhpSpec\Exception\Exception;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
+use Http\Message\Exception\EmptyStringException;
 
 /**
  * A formatter that prints a cURL command for HTTP requests.
- *
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
 class CurlCommandFormatter implements Formatter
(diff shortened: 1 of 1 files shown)


llaakkkk:
I create PR, but I have several variants of task solve, but I don’t know which is needed.
 Also, another solution can be by count NULL bytes in a string. What do you think about it?

src/Formatter/CurlCommandFormatter.php
```diff
@@ -45,7 +45,12 @@ public function formatRequest(RequestInterface $request)
             } else {
                 $data = '[non-seekable stream omitted]';
             }
-            $command .= sprintf(' --data %s', escapeshellarg($data));
+
+            $escapedData = @escapeshellarg($data);
+            if (isset($php_errormsg)) {
```
Nyholm: Wouldn't this always be false? Nothing defines `$php_errormsg`, right?

src/Formatter/CurlCommandFormatter.php
```diff
@@ -45,7 +45,12 @@ public function formatRequest(RequestInterface $request)
...
+                throw new \InvalidArgumentException($php_errormsg);
```
Nyholm: We do not want any exceptions or errors here. We want to silently fail. If we want to be extra nice to the user we may include a error message in the command. 

```php
// if fail:
$escapedData = 'We cound not escape the data properly';
```

Nyholm:
Any updates on this PR?

llaakkkk:
@Nyholm I'll do it asap

joelwurtz:
Thanks, you good for me, poke @Nyholm

PR #95, opened 2018-03-14
Fix implementation of __toString()

mxr576:
"This method MUST attempt to seek to the beginning of the stream before reading data and read the stream until the end is reached."

https://github.com/php-fig/http-message/blob/master/src/StreamInterface.php#L28

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no

#### What's in this PR?

This is a followup on this bug reported for the client-common package: https://github.com/php-http/client-common/issues/98

### DIFF
--- a/src/Encoding/FilteredStream.php
+++ b/src/Encoding/FilteredStream.php
@@ -154,6 +154,7 @@ abstract class FilteredStream implements StreamInterface
         if ($this->stream->isSeekable()) {
             $this->stream->rewind();
         }
+
         return $this->getContents();
     }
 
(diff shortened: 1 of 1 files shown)


xabbuh:
might be a good idea to even add a test for this

joelwurtz:
Thanks for this PR.

Problem is that it will always be false, we use a decorator here but we should have implemented directly this interface as a FilteredStream cannot be seekable.

As an example, a deflate stream will build it's deflate "table" (how to decode this string) on the fly, which means reseeking at the begin of the stream will produce a different response if using the same table. (You need to reset the table also)

IMO, the correct way would be to override `isSeekable`, `seek` and `rewind` functions to not allow those method (return false / throw exception), and an optional buffer option on top of plugin using the BufferedStream implementation.

joelwurtz:
We can close this if #100 is merged

dbu:
#100 has been merged

PR #100, opened 2018-07-09
Filtered Stream is not seekable

### LINKED ISSUE #90: Reading GzipDecodeStream causes Uncaught RuntimeException: Unable to perform operation on closed stream
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 1.0.1

#### Actual Behavior

What is the actual behavior?

`PHP Fatal error:  Uncaught RuntimeException: Unable to perform operation on closed stream in /Users/iconnor/Documents/PHP/php-http-message-bug/vendor/clue/stream-filter/src/functions.php:98`

#### Expected Behavior

What is the behavior you expect?

The stream is rewindable. I should be able to read its contents, rewind it, and then read its contents again.

#### Steps to Reproduce

Checkout this repository https://github.com/iainconnor/php-http-message-bug/blob/master/demo.php run demo.php.

Or,

``` php
$streamFactory = new \Http\Message\StreamFactory\GuzzleStreamFactory();
$stream = $streamFactory->createStream(file_get_contents('http://httpbin.org/gzip'));
$gzipStream = new \Http\Message\Encoding\GzipDecodeStream($stream);

echo $gzipStream->getContents() . PHP_EOL;

if ( $gzipStream->isSeekable() ) {
    $gzipStream->rewind();
    echo $gzipStream->getContents();
}
```

#### Possible Solutions

If you have already ideas how to solve the issue, add them here.
(remove this section if not needed)

These lines -- https://github.com/php-http/message/blob/master/src/Encoding/FilteredStream.php#L116-L118 -- look to be the cause, though I'm not sure on the solution.

joelwurtz:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | replace #95, fixes #90
| Documentation   | 
| License         | MIT

Filtered Stream is not seekable, previously we would rely on the inner stream, but it was a mistake.

### DIFF
--- a/src/Encoding/FilteredStream.php
+++ b/src/Encoding/FilteredStream.php
@@ -155,6 +155,30 @@ abstract class FilteredStream implements StreamInterface
     }
 
     /**
+     * {@inheritdoc}
+     */
+    public function isSeekable()
+    {
+        return false;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function rewind()
+    {
+        throw new \RuntimeException('Cannot rewind a filtered stream');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function seek($offset, $whence = SEEK_SET)
+    {
+        throw new \RuntimeException('Cannot seek a filtered stream');
+    }
+
+    /**
      * Returns the read filter name.
      *
      * @return string


joelwurtz:
Because the transformation of filtering stream is done with a "pipe" to ensure there is no memory problem, so nothing is keeped from the previous stream.

Making it seekable would force us to retain that 4 chars filtered equals to X chars not filtered and vice versa, and this is not done actually.

Nyholm:
Okey. 
So seek(2) would seek to the original stream's 2nd char but not necessarily the 2nd char of the actual (filtered) stream. 

Im 👍

joelwurtz:
Yes we need to sync both streams to make it works, and it's really difficult and prone to many errors, where using a BufferedStream is a really simple thing that can be add on top of this to allow seeking

GrahamCampbell:
Any news on this?

ste93cry:
This is currently a BC breaking change. Everyone extending the `FilteredStream` will now get an exception if he tries to rewind the stream, e.g. the build of [`sentry/sentry-php`](https://github.com/getsentry/sentry-php) is now failing

ste93cry:
I think that the best option would be to use `@trigger_error` to emit a deprecation warning and then fix the issue in the next major release. This will not introduce any BC in minor or patch versions

dbu:
i would consider returning `false` for `isSeekable` as acceptable change, and only replace the exceptions with the trigger_error calls.

sagikazarmark:
> oh, did it previously silently not work?

Well, as far as I can tell it at least returned something, that in some cases may have been correct, but using it is certainly unsafe.

Returning false would probably work if this class wasn't used as a parent class. Otherwise this change would be perfectly fine IMHO, since one should not rely on specific behaviour, but an interface instead which allows returning false and throwing an exception.

joelwurtz:
For me it's not BC since we strictly follow PSR7 here: you should always check if the stream is seekable if you want to rewind it. Since we are returning false, you should not use the rewind function, and use some sort of buffer.

dbu:
now i am confused ;-)
joel, do you want to say it *is* backwards compatible (BC) or it *is not* backwards compatible (aka a BC break)?

joelwurtz:
it is BC sorry my mistake, edited xD

dbu:
i think we are on a thin line of BC here, where things might have accidentally worked before (when rewind was not necessary). to avoid troubles for consumers, i would prefer to only keep the change to false, but `trigger_error` instead of throwing exceptions. this gives consumers that use the stream incorrectly time to fix their code.

sagikazarmark:
It looks like that would cause less harm at this point, but still leaves the previous issue open. :\

@joelwurtz would you mind opening a PR that reverts the change?

dbu:
see #104

dbu:
fixed in 1.7.2

joelwurtz:
Thanks @dbu

PR #101, opened 2018-07-09
Remove internal deprecation notice

### LINKED ISSUE #78: False positive trigger of deprecation notice in FilteredStream

joelwurtz:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #78 
| Documentation   | 
| License         | MIT

This remove internal deprecation notices and still respect bc promise, by implementing construct in each implementation.

### DIFF
--- a/src/Encoding/CompressStream.php
+++ b/src/Encoding/CompressStream.php
@@ -2,6 +2,7 @@
 
 namespace Http\Message\Encoding;
 
+use Clue\StreamFilter as Filter;
 use Psr\Http\Message\StreamInterface;
 
 /**
@@ -21,7 +22,10 @@ class CompressStream extends FilteredStream
             throw new \RuntimeException('The zlib extension must be enabled to use this stream');
         }
 
-        parent::__construct($stream, ['window' => 15, 'level' => $level], ['window' => 15]);
+        parent::__construct($stream, ['window' => 15, 'level' => $level]);
+
+        // @deprecated will be removed in 2.0
+        $this->writeFilterCallback = Filter\fun($this->writeFilter(), ['window' => 15]);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


Nyholm:
I see that this removes the deprecation warning when using ie `CompressStream`. If Im correct, a user that uses the `CompressStream` wil not notice any difference in 2.0, right? The `$writeFilterCallback` property is never used... anywhere.

joelwurtz:
Not sure how this was used, so i made this to be fully BC compatible, but we can remove setting this value if you want

Nyholm:
Im fine with the current implementation. I just wanted to make sure.

GrahamCampbell:
Any news on this?

dbu:
i have no experience with the php stream code, so i don't feel confident enough to merge this one. if people agree that this is the right way to do it, lets merge it.

PR #104, opened 2018-10-30
trigger_error instead of exceptions

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | reverts BC break of #100
| Deprecations?   | yes
| Related tickets | fixes #100 
| Documentation   | -
| License         | MIT

#### What's in this PR?

Use trigger_error rather than throw exceptions

#### Why?

Consumers SHOULD are supposed to check `isSeekable` and only seek/rewind if that returns true, but it seems they don't all do that. 

#### Checklist

### DIFF
--- a/src/Encoding/FilteredStream.php
+++ b/src/Encoding/FilteredStream.php
@@ -15,7 +15,10 @@ abstract class FilteredStream implements StreamInterface
 {
     const BUFFER_SIZE = 8192;
 
-    use StreamDecorator;
+    use StreamDecorator {
+        rewind as private doRewind;
+        seek as private doSeek;
+    }
 
     /**
      * @var callable
@@ -146,11 +149,11 @@ abstract class FilteredStream implements StreamInterface
     }
 
     /**
-     * {@inheritdoc}
+     * Always returns null because we can't tell the size of a stream when we filter.
      */
     public function getSize()
     {
-        return;
+        return null;
     }
 
     /**
@@ -162,7 +165,9 @@ abstract class FilteredStream implements StreamInterface
     }
 
     /**
-     * {@inheritdoc}
+     * Filtered streams are not seekable.
+     *
+     * We would need to buffer and process everything to allow seeking.
      */
     public function isSeekable()
     {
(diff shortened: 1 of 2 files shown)


src/Encoding/FilteredStream.php
```diff
@@ -174,15 +179,17 @@ public function isSeekable()
      */
     public function rewind()
     {
-        throw new \RuntimeException('Cannot rewind a filtered stream');
+        @trigger_error('Filtered streams are not seekable. This method will start raising an exception in the next major version');
+        $this->doRewind();
```
dbu: why is rewind not possible with a filtered stream? is the filter expected to track what it already saw and would need resetting? otherwise it seems to me that only `seek` is problematic, but not `rewind`

src/Encoding/FilteredStream.php
```diff
@@ -174,15 +179,17 @@ public function isSeekable()
      */
     public function rewind()
     {
-        throw new \RuntimeException('Cannot rewind a filtered stream');
+        @trigger_error('Filtered streams are not seekable. This method will start raising an exception in the next major version');
```
xabbuh: Maybe pass `E_USER_DEPRECATED` as the second argument to handle it as a normal deprecation?

src/Encoding/FilteredStream.php
joelwurtz: It really depends on the implementation, but let's say you have a filtered stream that hash the body, rewinding the underlying body would then certainly change the hash produced

Filtered stream act as a pipe, you should never go back.

Also according to PSR7 rewind should raise an exception if the stream is not seekable

src/Encoding/FilteredStream.php
joelwurtz: IMHO it's not a deprecation, but rather being nice to people using a buggy behavior

src/Encoding/FilteredStream.php
dbu: thanks for the clarification. so basically "you can't know" so you should not do it.

src/Encoding/FilteredStream.php
sagikazarmark: I agree with Joel's reasoning, but raising it as a deprecation might help when using tools that detect deprecations.

dbu: changed from default (E_USER_NOTICE) to E_USER_DEPRECATED for better visibility

src/Encoding/FilteredStream.php
joelwurtz: exactly

PR #106, opened 2018-12-26
Added deprecation notification on our factories

### LINKED ISSUE #105: PSR-17 compatibility ?
| Q            | A
| ------------ | ---
| Bug?         | no (not really)
| New Feature? | yes (kinda ?)
| Version      | latest

#### Actual Behavior
Seems that this package doesn't implement the `psr/http-factory` interfaces, and thus does not provide `psr/http-factory-implementation`.

#### Expected Behavior
To implement psr-17 interfaces instead of php-http's

#### Possible Solutions
Use the new interfaces from PSR-17, but as it would probably introduce BC breaks (as the psr are php 7.0+, and uses return typehints), it should be on a new major.

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | fixes #105 
| Documentation   | 
| License         | MIT

#### What's in this PR?

PSR-17 is out. Our factories are not BC compatible with PSR-17 because of PHP7 return types. I think we should remove our factories in version 2. 

Note. I do not suggest we plan for a version 2 release.

### DIFF
--- a/src/MessageFactory/DiactorosMessageFactory.php
+++ b/src/MessageFactory/DiactorosMessageFactory.php
@@ -11,6 +11,7 @@ use Zend\Diactoros\Response;
  * Creates Diactoros messages.
  *
  * @author GeLo <geloen.eric@gmail.com>
+ * @deprecated This will be removed in 2.0. Consider using the official Diactoros PSR-17 factory
  */
 final class DiactorosMessageFactory implements MessageFactory
 {
--- a/src/MessageFactory/DiactorosMessageFactory.php
+++ b/src/MessageFactory/DiactorosMessageFactory.php
@@ -12,7 +12,7 @@ use Zend\Diactoros\Response;
  *
  * @author GeLo <geloen.eric@gmail.com>
  *
- * @deprecated This will be removed in 2.0. Consider using the official Diactoros PSR-17 factory
+ * @deprecated This will be removed in php-http/message2.0. Consider using the official Diactoros PSR-17 factory
  */
 final class DiactorosMessageFactory implements MessageFactory
 {
--- a/src/MessageFactory/DiactorosMessageFactory.php
+++ b/src/MessageFactory/DiactorosMessageFactory.php
@@ -11,6 +11,7 @@ use Zend\Diactoros\Response;
  * Creates Diactoros messages.
  *
  * @author GeLo <geloen.eric@gmail.com>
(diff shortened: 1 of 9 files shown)


dbu:
should we additionally deprecate our factory interfaces? i'd say so.

also, symfony does a `trigger_error('...', E_USER_DEPRECATED);` to make the problem more visible. should we do that here too?

Nyholm:
I think we should deprecate our interfaces yes. 

About trigger_error. I was considering that. I don't think it is needed. Since we currently have not plan of ever release a new major version I don't think we need to make it very visible. This annotation will show that you should never create a *new* implementation with these factories

dbu:
agreed, so lets not trigger. can you please update this with deprecation on the interface and fix the code style issue (blank line around `@deprecated` line) ?

Nyholm:
I fixed the code style issues. 

Here is a PR for the interfaces: https://github.com/php-http/message-factory/pull/39

gmponos:
There is also one more reason not to have trigger_error and that is that libraries like guzzle do not have factories implemented yet.

Otherwise my opinion would've been that it would be better to raise the error.

sagikazarmark:
I would say this PR should be the last one merged and we could even wait for guzzle to release PSR-17

src/MessageFactory/DiactorosMessageFactory.php
```diff
@@ -11,6 +11,8 @@
  * Creates Diactoros messages.
  *
  * @author GeLo <geloen.eric@gmail.com>
+ *
+ * @deprecated This will be removed in 2.0. Consider using the official Diactoros PSR-17 factory
```
xabbuh: We may want to include the package name together with the version. This way, it will be more obvious which package deprecated the interface (the Symfony Debug component turns the `@deprecated` annotation into `E_USER_DEPRECATED` errors IIRC).

dbu:
would this now be ready to merge? is guzzle psr-17 released?

dbu:
@Nyholm what is the state of this?

dbu:
@Nyholm do we want to merge this?

Nyholm:
Yes, Im 👍 for merging. 

I've updated the warning and rebased. The CS issues are unrelated.

PR #107, opened 2018-12-26
Test on php 7.3

Nyholm:


joelwurtz:
From what i have seen it seems that `stream_set_chunk_size` does not work like before in php 7.2. Don't know what change but this is only explanation possible.

joelwurtz:
Ok, after some search it's coming from fread which now better respect length pass to it #110 should fix those tests

dbu:
can you please rebase on master now that the chunk issue should be fixed?

joelwurtz:
I think tests are played on the merged version and since there are green we can safely merge this ?

dbu:
thanks. indeed, restarting the tests was enough.

PR #110, opened 2019-02-17
Always read size of chunksize if in MemoryStream

joelwurtz:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #107
| Documentation   | /
| License         | MIT

This fix test for php 7.2 / 7.3, PHP seems now to respect fread length better since those version exists, which is misleading in our test since we try to read the default php buffer size (8192)

### DIFF
--- a/spec/Encoding/MemoryStream.php
+++ b/spec/Encoding/MemoryStream.php
@@ -10,7 +10,9 @@ class MemoryStream implements StreamInterface
 
     private $size = 0;
 
-    public function __construct($body = "", $chunkSize = null)
+    private $chunkSize;
+
+    public function __construct($body = "", $chunkSize = 8192)
     {
         $this->size = strlen($body);
         $this->resource = fopen('php://memory', 'rw+');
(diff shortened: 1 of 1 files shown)


spec/Encoding/MemoryStream.php
```diff
@@ -10,7 +10,9 @@ class MemoryStream implements StreamInterface
 
     private $size = 0;
 
-    public function __construct($body = "", $chunkSize = null)
+    private $chunkSize;
+
+    public function __construct($body = "", $chunkSize = 8192)
```
dbu: how did you determine this default? is there a way to figure out the system default instead of hardcoding it here? i guess 8kb is the norm for 64bit and 4096 for 32bit?

joelwurtz: Default value is the `BUFSIZ` C constant which depends on the compiler used, and there is no way to get it here.

When reading various streaming PHP librairies, 8192 was a widely value used for buffer, so decided to keep here.

dbu: thanks for the explanation!
oh this is the spec and not the source. sorry i asked, for a test we don't need to overthink it anyways.

PR #116, opened 2019-08-02
Fix fatal error for long bodies

### LINKED ISSUE #93: CurlFormatter fails on large boddies

GaryPEGEOT:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #93 
| Documentation   | 
| License         | MIT

#### What's in this PR?

Prevent `CurlCommandFormatter` to throw a Fatal exception when body size exceed `escapeshellarg` argument allowed length. Didn't find a great way to get the limit tho... Windows limit come from [PHP source code](https://github.com/php/php-src/blob/959091285487878790f2f2fe71301954f3c765a9/ext/standard/exec.c#L62-L80), the UNIX from 
```
echo $(getconf ARG_MAX)
```
If you have a better solution to find `escapeshellarg` limit, I'm all ears

#### Checklist

- [x] Updated CHANGELOG.md to describe bugfix

### DIFF
--- a/src/Formatter/CurlCommandFormatter.php
+++ b/src/Formatter/CurlCommandFormatter.php
@@ -36,7 +36,7 @@ class CurlCommandFormatter implements Formatter
 
         $body = $request->getBody();
         if ($body->getSize() > 0) {
-            // escapeshellarg argument max length on Windows, but impractical to use either way
+            // escapeshellarg argument max length on Windows, but longer body in curl command would be impractical anyways
             if ($body->getSize() > 8192) {
                 $data = '[too long stream omitted]';
             } elseif ($body->isSeekable()) {
--- a/src/Formatter/CurlCommandFormatter.php
+++ b/src/Formatter/CurlCommandFormatter.php
@@ -36,10 +36,8 @@ class CurlCommandFormatter implements Formatter
 
         $body = $request->getBody();
         if ($body->getSize() > 0) {
-            // escapeshellarg argument max length
-            $argMaxLength = '\\' === DIRECTORY_SEPARATOR ? 8192 : 2097152;
-
-            if ($body->getSize() > $argMaxLength) {
+            // escapeshellarg argument max length on Windows, but impractical to use either way
+            if ($body->getSize() > 8192) {
                 $data = '[too long stream omitted]';
             } elseif ($body->isSeekable()) {
                 $data = $body->__toString();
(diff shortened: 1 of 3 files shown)


GaryPEGEOT:
> maybe lets just limit to 8192 and say that this should work on all operating systems. wdyt?

Totally agree, it's neither user friendly nor memory friendly

src/Formatter/CurlCommandFormatter.php
```diff
@@ -36,7 +36,10 @@ public function formatRequest(RequestInterface $request)
 
         $body = $request->getBody();
         if ($body->getSize() > 0) {
-            if ($body->isSeekable()) {
+            // escapeshellarg argument max length on Windows, but impractical to use either way
```
dbu: maybe say "... max length on Windows, but longer body in curl command would be impractical anyways."

src/Formatter/CurlCommandFormatter.php
```diff
@@ -36,7 +36,10 @@ public function formatRequest(RequestInterface $request)
...
+            if ($body->getSize() > 8192) {
+                $data = '[too long stream omitted]';
+            } elseif ($body->isSeekable()) {
                 $data = $body->__toString();
```
dbu: could we have a body that does not know its size? then we might need to check again how long $data is?

GaryPEGEOT: on line 38 your have 
```php
if ($body->getSize() > 0) {
```

So I guess passing that point, it knows it size. But I can add another check just in case

src/Formatter/CurlCommandFormatter.php
GaryPEGEOT: Updated :)

src/Formatter/CurlCommandFormatter.php
dbu: oh, i missed that. no then its fine!

dbu:
thanks!

dbu:
released 1.8.0 with this and some other fixes.

PR #121, opened 2019-12-27
Enhancement: Use ergebnis/composer-normalize instead of localheinz/composer-normalize

localheinz:
This PR

* [x] uses `ergebnis/composer-normalize` instead of `localheinz/composer-normalize`

Related to https://github.com/ergebnis/composer-normalize/issues/266.

💁‍♂ For reference, see https://localheinz.com/blog/2019/12/10/from-localheinz-to-ergebnis/.


localheinz:
Thank you, @sagikazarmark!

PR #119, opened 2019-12-16
Fix CS

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

PR #122, opened 2019-12-30
Add GitHub actions

sagikazarmark:
Closes #119

PR #123, opened 2020-01-02
disable a new symfony rule that we don't want to follow

dbu:
fix #119


dbu:
need to check when #122 is merged, as styleci has been disabled.

sagikazarmark:
I already disabled this rule in #122. I would rather use GitHub actions and/or PrettyCI.

dbu:
yep, that makes sense.

PR #133, opened 2020-10-12
fixed detection of binary strings (#132)

### LINKED ISSUE #132: Binary string detection is broken (introduced with 1.9.0)
**PHP version**: all versions

**Description**
With #126 and #128 a change was introduced that is supposed to prevent binary messages from being dumped. In [FullHttpMessageFormatter](https://github.com/php-http/message/blob/master/src/Formatter/FullHttpMessageFormatter.php#L89) line 89 the following code is used to detect binary strings:

```php
        if (preg_match('/([\x00-\x1F\x7F])/', $data)) {
            return $message.'[binary stream omitted]';
        }
```

Unfortunatly, this regex includes common whitespaces like \t, \r and \n as well. In consequence, all output is hidden and replaced with `[binary stream omitted]`, even if the document is a simple HTML or XML document. I doubt that this is the intentioned behaviour, since linebreaks a by far an indicator for a binary string and moreover very common in HTML or XML documents.

**Possible Solution**
The code could be changed as follows, see https://stackoverflow.com/questions/25343508/detect-if-string-is-binary:

```php
        if (preg_match('~[^\x20-\x7E\t\r\n]~', $data)) {
            return $message.'[binary stream omitted]';
        }
```

However, this would still be problematic regarding non-English languages, i.e. special characters in French, Spanish, Italian or German would still trigger falsely the binary detection. Therefore, I would suggest to check for all non-prinable ASCII characters as well as \<del> (\x7F) and explicitly exclude \<cr> (\x0D), \<lf> (\x0A) and \<tab> (\x0B).

```php
        // all non-printable ASCII characters and <DEL> except for \t, \r, \n
        if (preg_match('/([\x00-\x09\x0C\x0E-\x1F\x7F])/', $data)) {
            return $message.'[binary stream omitted]';
        }
```

The same change should probably be applied to [CurlCommandFormatter](https://github.com/php-http/message/blob/master/src/Formatter/CurlCommandFormatter.php) as well.

What do you think?

driehle:
| Q               | A
| --------------- | ---
| Bug fix?         | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #132
| License         | MIT

See #132, tests have been added.

### DIFF
--- a/spec/Formatter/CurlCommandFormatterSpec.php
+++ b/spec/Formatter/CurlCommandFormatterSpec.php
@@ -91,6 +91,24 @@ class CurlCommandFormatterSpec extends ObjectBehavior
         $this->formatRequest($request)->shouldReturn("curl 'http://foo.com/bar' --request POST --data '[binary stream omitted]'");
     }
 
+    function it_formats_requests_with_line_break(RequestInterface $request, UriInterface $uri, StreamInterface $body)
+    {
+        $request->getUri()->willReturn($uri);
+        $request->getBody()->willReturn($body);
+
+        $body->__toString()->willReturn("foo\nbar");
+        $body->getSize()->willReturn(1);
+        $body->isSeekable()->willReturn(true);
+        $body->rewind()->willReturn(true);
+
+        $uri->withFragment('')->willReturn('http://foo.com/bar');
+        $request->getMethod()->willReturn('POST');
+        $request->getProtocolVersion()->willReturn('1.1');
+        $request->getHeaders()->willReturn([]);
+
+        $this->formatRequest($request)->shouldReturn("curl 'http://foo.com/bar' --request POST --data 'foo\nbar'");
+    }
+
     function it_formats_requests_with_nonseekable_body(RequestInterface $request, UriInterface $uri, StreamInterface $body)
     {
         $request->getUri()->willReturn($uri);
--- a/src/Formatter/CurlCommandFormatter.php
+++ b/src/Formatter/CurlCommandFormatter.php
@@ -42,7 +42,8 @@ class CurlCommandFormatter implements Formatter
             } elseif ($body->isSeekable()) {
                 $data = $body->__toString();
                 $body->rewind();
-                if (preg_match('/[\x00-\x1F\x7F]/', $data)) {
+                // all non-printable ASCII characters and <DEL> except for \t, \r, \n
+                if (preg_match('/([\x00-\x09\x0C\x0E-\x1F\x7F])/', $data)) {
                     $data = '[binary stream omitted]';
                 }
             } else {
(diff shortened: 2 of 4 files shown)


src/Formatter/CurlCommandFormatter.php
```diff
@@ -42,9 +42,12 @@ public function formatRequest(RequestInterface $request)
             } elseif ($body->isSeekable()) {
                 $data = $body->__toString();
                 $body->rewind();
-                if (preg_match('/[\x00-\x1F\x7F]/', $data)) {
+                // all non-printable ASCII characters and <DEL> except for \t, \r, \n
+                if (preg_match('/([\x00-\x09\x0C\x0E-\x1F\x7F])/', $data)) {
                     $data = '[binary stream omitted]';
                 }
+                // replace \t, \r and \n with their string literals
+                $data = str_replace(["\t", "\r", "\n"], ['\t', '\r', '\n'], $data);
```
dbu, line 50: if i copy-paste this onto a console, does the `\t` / `\r` / `\n` work correctly? i am a bit afraid of cluttering the content with such characters - but when we output as html, the newlines would get lost, so this might be the right solution.

driehle: Good catch, it will not work due to the `escapeshellarg()` later on. We could simple not care about these three characters at all and `escapeshelarg()` will do the rest. The output will then look somehow like this:

```
curl --......  --data 'foo
bar' http://foo/bar
```

This doesn't look nice, but it can be copy and pasted.

dbu: i prefer less nice but working :-D

spec/Formatter/CurlCommandFormatterSpec.php
```diff
@@ -91,6 +91,24 @@ function it_formats_requests_with_null_bytes(RequestInterface $request, UriInter
...
+        $body->getSize()->willReturn(1);
+        $body->isSeekable()->willReturn(true);
+        $body->rewind()->willReturn(true);
+
+        $uri->withFragment('')->willReturn('http://foo.com/bar');
+        $request->getMethod()->willReturn('POST');
+        $request->getProtocolVersion()->willReturn('1.1');
+        $request->getHeaders()->willReturn([]);
+
+        $this->formatRequest($request)->shouldReturn("curl 'http://foo.com/bar' --request POST --data 'foo\\nbar'");
```
dbu, line 109:
```suggestion
        $this->formatRequest($request)->shouldReturn("curl 'http://foo.com/bar' --request POST --data 'foo\nbar'");
```

dbu:
thanks a lot, i will tag a release

Issue #125, opened 2020-01-09
Package zendframework/zend-diactoros is abandoned


PVGrad:
Composer warns:

>Package zendframework/zend-diactoros is abandoned, you should avoid using it. Use laminas/laminas-diactoros instead.

See: https://www.zend.com/blog/what-status-zend-framework-transition-laminas

dbu:
for BC reasons, i think we need to keep the zend factories and such for now.

if you want to contribute a laminas factory, please do. you could also mark the zend factory as deprecated.

driehle:
@dbu You can safely migrate from zend/zend-diactoros to laminas/laminas-diactoros. Doing so will, as a dependency, automatically require the package laminas/laminas-zendframework-bridge. This package provides a mechanism which intercepts all class loading requests to legacy classes (Zend\ namespace), load the new class and create a class alias. See https://github.com/laminas/laminas-zendframework-bridge/blob/1.2.x/src/Autoloader.php#L112 for the actual implementation. Therefore, you won't have a BC break due to the migration. I'll provide a PR.

dbu:
fixed in #134

PR #134, opened 2020-10-12
migrated to laminas/laminas-diactoros (#125)

driehle:
| Q               | A
| --------------- | ---
| Bug fix?         | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #125 
| License         | MIT

#### What's in this PR?

This PR replaces zend/zend-diactoros with laminas/laminas-diactoros. Due to laminas/laminas-zendframework-bridge, there is no BC break.

#### Why?

Because zend/zend-diactoros is officially abandoned.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,13 +29,13 @@
         "guzzlehttp/psr7": "^1.0",
         "phpspec/phpspec": "^5.1 || ^6.3",
         "slim/slim": "^3.0",
-        "zendframework/zend-diactoros": "^1.0"
+        "laminas/laminas-diactoros": "^2.0"
     },
     "suggest": {
         "ext-zlib": "Used with compressor/decompressor streams",
         "guzzlehttp/psr7": "Used with Guzzle PSR-7 Factories",
-        "slim/slim": "Used with Slim Framework PSR-7 implementation",
-        "zendframework/zend-diactoros": "Used with Diactoros Factories"
+        "laminas/laminas-diactoros": "Used with Diactoros Factories",
+        "slim/slim": "Used with Slim Framework PSR-7 implementation"
     },
     "config": {
         "sort-packages": true
--- a/puli.json
+++ b/puli.json
@@ -7,7 +7,7 @@
             "class": "Http\\Message\\MessageFactory\\DiactorosMessageFactory",
             "type": "Http\\Message\\MessageFactory",
             "parameters": {
-                "depends": "Zend\\Diactoros\\Request"
+                "depends": "Laminas\\Diactoros\\Request"
             }
         },
         "0836751e-6558-4d1b-8993-4a52012947c3": {
@@ -57,7 +57,7 @@
             "class": "Http\\Message\\UriFactory\\DiactorosUriFactory",
             "type": "Http\\Message\\UriFactory",
             "parameters": {
-                "depends": "Zend\\Diactoros\\Uri"
+                "depends": "Laminas\\Diactoros\\Uri"
             }
         },
         "4672a6ee-ad9e-4109-a5d1-b7d46f26c7a1": {
@@ -70,7 +70,7 @@
             "class": "Http\\Message\\MessageFactory\\DiactorosMessageFactory",
             "type": "Http\\Message\\ResponseFactory",
             "parameters": {
-                "depends": "Zend\\Diactoros\\Response"
+                "depends": "Laminas\\Diactoros\\Response"
             }
         },
         "6a9ad6ce-d82c-470f-8e30-60f21d9d95bf": {
@@ -88,7 +88,7 @@
             "class": "Http\\Message\\StreamFactory\\DiactorosStreamFactory",
             "type": "Http\\Message\\StreamFactory",
             "parameters": {
-                "depends": "Zend\\Diactoros\\Stream"
+                "depends": "Laminas\\Diactoros\\Stream"
             }
         },
         "a018af27-7590-4dcf-83a1-497f95604cd6": {
@@ -104,7 +104,7 @@
             "class": "Http\\Message\\MessageFactory\\DiactorosMessageFactory",
             "type": "Http\\Message\\RequestFactory",
             "parameters": {
-                "depends": "Zend\\Diactoros\\Request"
+                "depends": "Laminas\\Diactoros\\Request"
             }
         }
     }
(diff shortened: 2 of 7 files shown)


puli.json
```diff
@@ -22,7 +22,7 @@
         },
         "2438c2d0-0658-441f-8855-ddaf0f87d54d": {
             "_class": "Puli\\Discovery\\Binding\\ClassBinding",
-            "class": "Http\\Message\\MessageFactory\\GuzzleMessageFactory",
+            "class": "Http\\Message\\MessageFactory\\GuzcuzleMessageFactory",
```
dbu, line 25: was this change on purpose?

driehle: No, I'll fix it

dbu: thanks

composer.json
```diff
@@ -30,15 +30,16 @@
         "ergebnis/composer-normalize": "^2.1",
         "guzzlehttp/psr7": "^1.0",
         "henrikbjorn/phpspec-code-coverage": "^1.0",
+        "laminas/laminas-diactoros": "^2.0",
+        "laminas/laminas-zendframework-bridge": "^1.0",
```
dbu, line 34: this is in the require-dev section.

consumers will use this library together with the implementation of their choice. i fear that we need to do this differently. we could either in the factory check if the laminas classes exist and otherwise fallback to zend. or we could leave the existing factories alone and create new LaminasDiactoros...Factory classes.

driehle: Well, the end user would need to do `composer require laminas/laminas-diactoros`. With that, the end user will not have to change anything in their code. The only thing that will not work is if only the legacy package zend/zend-diactoros is required, but not laminas/lamians-diactoros. Don't you think it would be fine to mention this in the docs?

driehle: If you prefer end users not to switch the library name, we could to:
```php
use Laminas\Diactoros\Request as LaminasRequest;
use Zend\Diactoros\Request as ZendRequest;

public function createRequest(...) 
{
    // this call will trigger autoload!
    if (class_exists('Laminas\Diactoros\Request')) {
        return new LaminasRequest(...);
    }

    return new ZendRequest(...);
}
```

dbu: the problem is, if we don't do a new major version, users might just `composer upgrade` and ignore the warning about zend being abandoned. composer will install, but they application is suddenly broken.

so i think we we should do your suggestion. does laminas always provide the BC layer of extending the old zend classes, or only if the additional bc package is installed?

dbu:
if i understand correctly, this means that if one installs only laminas, but not the zend-bridge, the factory now returns a different class. i would hope that users do not rely on the concrete implementations when using this factory (otherwise there is no need to use the factory anyways). so to me this seems fine. 

i suggest we do a new minor version, as it still is more than a bugfix. can you please rebase on the master branch and add something to the changelog (including mentioning the bridge?)

composer.json
dbu, line 34: why install the bridge for the tests? things would work fine without, right?

driehle: Yes, we could drop the bridge here.

driehle:
> if i understand correctly, this means that if one installs only laminas, but not the zend-bridge, the factory now returns a different class. i would hope that users do not rely on the concrete implementations when using this factory (otherwise there is no need to use the factory anyways). so to me this seems fine.

Yes, you need to have the bridge installed. However, the bridge is required as a dependency of laminas-diactoros (see https://packagist.org/packages/laminas/laminas-diactoros) and will, to my understanding, be included there as a dependency until laminas-diactoros releases a major update (i.e. 3.0). So as soon as someone will have laminas-diactoros 2.x installed, they will have the bridge installed as well.

In the [autoloader](https://github.com/laminas/laminas-zendframework-bridge/blob/1.2.x/src/Autoloader.php#L112) of the bridge you can see that they use [class_alias()](https://www.php.net/class_alias) to create the legacy classes (e.g. ZendRequest). Therefore, if you have laminas-diactoros and the bridge installed, you will actually use LaminasRequest, but an $laminasRequest instanceof ZendRequest will still hold true (and vice versa, tough that is unlikely to happen).

> i suggest we do a new minor version, as it still is more than a bugfix. can you please rebase on the master branch and add something to the changelog (including mentioning the bridge?)

Sure, will do.

driehle:
Rebased and changelog updated. I fixed some broken and missing links in the changelog as well. ;-)

puli.json
```diff
@@ -7,7 +7,7 @@
             "class": "Http\\Message\\MessageFactory\\DiactorosMessageFactory",
             "type": "Http\\Message\\MessageFactory",
             "parameters": {
-                "depends": "Zend\\Diactoros\\Request"
+                "depends": "Laminas\\Diactoros\\Request"
```
dbu, line 10: not sure how many people use puli, but i think we should only add the laminas things but not replace zend.

driehle: I am not using puli and also not sure about its usefullness in general, but the way in understand this would be that adding two dependencies would mean that both (AND) are required and not only one (OR). If this is the case, then we should leave it with laminas here.

dbu: @Nyholm do you know if we can make puli detect both the old and the new class?

if we can't i opt to configure it with the old class. when the BC layer is there, that should even still work.

dbu: @Nyholm wdyt?

dbu:
can you please fix the cs issue? https://github.com/php-http/message/pull/134/checks?check_run_id=1246548559

if tobias can clarify the puli usage, we can adjust that, if not i suggest we revert the puli change to not introduce unexpected BC breaks.

driehle:
> can you please fix the cs issue? https://github.com/php-http/message/pull/134/checks?check_run_id=1246548559
> 
> if tobias can clarify the puli usage, we can adjust that, if not i suggest we revert the puli change to not introduce unexpected BC breaks.

@dbu: Rebased on current master, fixed CS issue. Still awaiting the reply of @Nyholm regarding puli.

puli.json
Nyholm: Im honestly not sure either.. :/

driehle: Ok, then we should probably stick to the old class name...

dbu (approved):
thank you!

PR #136, opened 2020-12-10
Resolves issue #43

### LINKED ISSUE #43: WSSE Authentication allow to change hashing algorithm
| Q | A |
| --- | --- |
| Bug? | no |
| New Feature? | yes |
#### Actual Behavior

The authentication uses SHA1 to hash the token.
#### Expected Behavior

User should be allowed to change hashing mechanism based on the server.
#### Possible Solutions

Add a new parameter in the constructor with the default value SHA1. The default value should be deprecated right away and be replaced with a mandatory one: check the number of arguments and trigger a deprecated error if we see the default value is used.

peter279k:
| Q               | A
| --------------- | ---
| Bug fix?         | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | issue #43
| Documentation   | N/A
| License         | MIT

Closes #43.

#### What's in this PR?

- Using the `hash('sha512')` to replace `sha1` function.

#### Checklist

- [ ] Documentation pull request created (if not simply a bugfix)

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -6,6 +6,7 @@ $config = PhpCsFixer\Config::create()
         '@Symfony' => true,
         'array_syntax' => ['syntax' => 'short'],
         'single_line_throw' => false,
+        'protected_to_private' => false,
     ])
     ->setFinder(
         PhpCsFixer\Finder::create()
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -6,7 +6,6 @@ $config = PhpCsFixer\Config::create()
         '@Symfony' => true,
         'array_syntax' => ['syntax' => 'short'],
         'single_line_throw' => false,
-        'protected_to_private' => false,
     ])
     ->setFinder(
         PhpCsFixer\Finder::create()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
    the respective classes are available via autoloading, but continue to return objects from `Zend\Diactoros\`
    namespace otherwise.
 
+## [1.11.0] - unreleased
+
+- Allow to specify the hashing algorithm for WSSE authentication.
+
 ## [1.10.0] - 2020-11-11
 
 - Added support for PHP 8.0.
--- a/src/Authentication/Wsse.php
+++ b/src/Authentication/Wsse.php
@@ -29,11 +29,6 @@ final class Wsse implements Authentication
     private $hashAlgorithm;
 
     /**
-     * @var array<string>
-     */
-    private $acceptedHashAlgos = ['sha1', 'sha512', 'sha384', 'sha3-384', 'sha3-512'];
-
-    /**
      * @param string $username
      * @param string $password
      * @param string $hashAlgorithm To use a better hashing algorithm than the weak sha1, pass the algorithm to use, e.g. "sha512"
@@ -42,6 +37,9 @@ final class Wsse implements Authentication
     {
         $this->username = $username;
         $this->password = $password;
+        if (false === in_array($hashAlgorithm, hash_algos())) {
+            throw new InvalidArgumentException(sprintf('Unaccepted hashing algorithm: %s', $hashAlgorithm));
+        }
         $this->hashAlgorithm = $hashAlgorithm;
     }
 
@@ -52,9 +50,6 @@ final class Wsse implements Authentication
     {
         $nonce = substr(md5(uniqid(uniqid().'_', true)), 0, 16);
         $created = date('c');
-        if (false === in_array($this->hashAlgorithm, $this->acceptedHashAlgos)) {
-            throw new InvalidArgumentException(sprintf('Unaccepted hashing algorithm: %s', $this->hashAlgorithm));
-        }
         $digest = base64_encode(hash($this->hashAlgorithm, base64_decode($nonce).$created.$this->password, true));
 
         $wsse = sprintf(
--- a/src/Authentication/Wsse.php
+++ b/src/Authentication/Wsse.php
@@ -3,6 +3,7 @@
 namespace Http\Message\Authentication;
 
 use Http\Message\Authentication;
+use InvalidArgumentException;
 use Psr\Http\Message\RequestInterface;
 
 /**
@@ -23,13 +24,25 @@ final class Wsse implements Authentication
     private $password;
 
     /**
+     * @var string
+     */
+    private $hashAlgorithm;
+
+    /**
+     * @var array<string>
+     */
+    private $acceptedHashAlgos = ['sha1', 'sha512', 'sha384', 'sha3-384', 'sha3-512'];
+
+    /**
      * @param string $username
      * @param string $password
+     * @param string $hashAlgorithm To use a better hashing algorithm than the weak sha1, pass the algorithm to use, e.g. "sha512"
      */
-    public function __construct($username, $password)
+    public function __construct($username, $password, $hashAlgorithm = 'sha1')
     {
         $this->username = $username;
         $this->password = $password;
+        $this->hashAlgorithm = $hashAlgorithm;
     }
 
     /**
@@ -37,10 +50,12 @@ final class Wsse implements Authentication
      */
     public function authenticate(RequestInterface $request)
     {
-        // TODO: generate better nonce?
         $nonce = substr(md5(uniqid(uniqid().'_', true)), 0, 16);
         $created = date('c');
-        $digest = base64_encode(sha1(base64_decode($nonce).$created.$this->password, true));
+        if (false === in_array($this->hashAlgorithm, $this->acceptedHashAlgos)) {
+            throw new InvalidArgumentException(sprintf('Unaccepted hashing algorithm: %s', $this->hashAlgorithm));
+        }
+        $digest = base64_encode(hash($this->hashAlgorithm, base64_decode($nonce).$created.$this->password, true));
 
         $wsse = sprintf(
             'UsernameToken Username="%s", PasswordDigest="%s", Nonce="%s", Created="%s"',
--- a/src/CookieJar.php
+++ b/src/CookieJar.php
@@ -10,7 +10,7 @@ namespace Http\Message;
 final class CookieJar implements \Countable, \IteratorAggregate
 {
     /**
-     * @var \SplObjectStorage
+     * @var \SplObjectStorage<object, mixed>
      */
     protected $cookies;
 
--- a/src/Encoding/FilteredStream.php
+++ b/src/Encoding/FilteredStream.php
@@ -13,12 +13,11 @@ use Psr\Http\Message\StreamInterface;
  */
 abstract class FilteredStream implements StreamInterface
 {
-    const BUFFER_SIZE = 8192;
-
     use StreamDecorator {
         rewind as private doRewind;
         seek as private doSeek;
     }
+    const BUFFER_SIZE = 8192;
 
     /**
      * @var callable


src/CookieJar.php
```diff
@@ -12,7 +12,7 @@ final class CookieJar implements \Countable, \IteratorAggregate
     /**
      * @var \SplObjectStorage
      */
-    protected $cookies;
+    private $cookies;
```
peter279k, line 15: Fixing coding style via `PHP-CS-Fixer`.

src/CookieJar.php
```diff
@@ -92,7 +92,7 @@ public function getMatchingCookies(Cookie $cookie)
      *
      * @return Cookie[]
      */
-    protected function findMatchingCookies(callable $match)
+    private function findMatchingCookies(callable $match)
```
peter279k, line 95: Fixing coding style via `PHP-CS-Fixer`.

src/CookieJar.php
peter279k: The coding style issue is as follows:

````
PHP CS Fixer 2.17.1 Desert Beast by Fabien Potencier and Dariusz Ruminski (5198b73)
Loaded config default from "/github/workspace/.php_cs.dist".
   1) src/CookieJar.php
      ---------- begin diff ----------
--- Original
+++ New
@@ -12,7 +12,7 @@
     /**
      * @var \SplObjectStorage
      */
-    protected $cookies;
+    private $cookies;
 
     public function __construct()
     {
@@ -92,7 +92,7 @@
      *
      * @return Cookie[]
      */
-    protected function findMatchingCookies(callable $match)
+    private function findMatchingCookies(callable $match)
     {
         $cookies = [];
 

      ----------- end diff -----------


Checked all files in 0.659 seconds, 12.000 MB memory used
````

Comment (dbu) [reply src/CookieJar.php:15]: i was worried for a bit, but then noticed that the class is final, so indeed having protected properties/methods makes no sense.

File: CHANGELOG.md [hunk 3]
```diff
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
    the respective classes are available via autoloading, but continue to return objects from `Zend\Diactoros\`
    namespace otherwise.
 
+## [1.11.0] - 2020-12-11
```
dbu, line 20:
```suggestion
## [1.11.0] - unreleased
```

composer.json
```diff
@@ -17,6 +17,7 @@
     "require": {
         "php": "^7.1 || ^8.0",
         "clue/stream-filter": "^1.5",
+        "friends-of-phpspec/phpspec-code-coverage": "^4.3.2",
```
dbu, line 20: please remove this from the require section.

src/Authentication/Wsse.php
```diff
@@ -40,7 +40,7 @@ public function authenticate(RequestInterface $request)
         // TODO: generate better nonce?
```
dbu, line 40: i think with this PR, we can remove the todo, no?

src/Authentication/Wsse.php
```diff
@@ -40,7 +40,7 @@ public function authenticate(RequestInterface $request)
...
         $nonce = substr(md5(uniqid(uniqid().'_', true)), 0, 16);
         $created = date('c');
-        $digest = base64_encode(sha1(base64_decode($nonce).$created.$this->password, true));
+        $digest = base64_encode(hash('sha512', base64_decode($nonce).$created.$this->password, true));
```
dbu, line 43: to fully resolve #43, there should be an optional constructor argument on this class for the algorithm that defaults to the algo we want to use.

i am not sure if we should change the default algorithm - if somebody is using this library with a server that only supports sha1 but not sha512, this would be a BC break for them. in a new major version, we can change to default to a state-of-the-art hash and let people overwrite to sha1 when needed...

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+
+- For issue #43.
```
dbu, line 22:
```suggestion
- Allow to specify the hasing algorithm for WSSE authentication.
```

dbu (changes requested):
thanks a lot for picking this one up!

i added some suggestions. @sagikazarmark do we need to worry about a BC break when the default changes from sha1 to sha512?

.php_cs.dist
```diff
@@ -6,6 +6,7 @@ $config = PhpCsFixer\Config::create()
         '@Symfony' => true,
         'array_syntax' => ['syntax' => 'short'],
         'single_line_throw' => false,
+        'protected_to_private' => false,
```
peter279k, line 9: Adding this setting to ignore final class with private properties checking.

src/Authentication/Wsse.php
```diff
@@ -23,24 +23,33 @@ final class Wsse implements Authentication
...
 
     /**
      * {@inheritdoc}
      */
     public function authenticate(RequestInterface $request)
     {
-        // TODO: generate better nonce?
         $nonce = substr(md5(uniqid(uniqid().'_', true)), 0, 16);
         $created = date('c');
         $digest = base64_encode(sha1(base64_decode($nonce).$created.$this->password, true));
```
dbu, line 49: i think this should be an if/else to not unnecessarly do the sha1 encoding and then throw it away immediately

src/Authentication/Wsse.php
```diff
@@ -23,24 +23,33 @@ final class Wsse implements Authentication
...
+        if (true === $this->useSha512) {
+            $digest = base64_encode(hash('sha512', base64_decode($nonce).$created.$this->password, true));
```
dbu, line 51: as the argument to `hash` is a string, could we make the constructor argument generic to pass whatever hashing algorithm is desired by the user? the useSha512 flag will become cumbersome if some point in the future there is sha1024 or whatever. if we just pass a string, we have maximum flexibility. we can phpdoc the $hashAlgorithm parameter to say `To use a better hashing algorithm than the weak sha1, pass the algorithm to use, e.g. "sha512"`

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
    the respective classes are available via autoloading, but continue to return objects from `Zend\Diactoros\`
    namespace otherwise.
 
+## [1.11.0] - unreleased
+
+- Allow to specify the hasing algorithm for WSSE authentication.
```
dbu, line 22:
```suggestion
- Allow to specify the hashing algorithm for WSSE authentication.
```

dbu (commented):
finally got around to look at this. can we make the hashing completely configurable, while we are at it?

.php_cs.dist
dbu: is this about some other class? because Wsse has no protected properties/methods.

peter279k: This setting is for following coding style checking issue:

```
Loaded config default from "/github/workspace/.php_cs.dist".
   1) src/CookieJar.php
      ---------- begin diff ----------
--- Original
+++ New
@@ -12,7 +12,7 @@
     /**
      * @var \SplObjectStorage
      */
-    protected $cookies;
+    private $cookies;
 
     public function __construct()
     {
@@ -92,7 +92,7 @@
      *
      * @return Cookie[]
      */
-    protected function findMatchingCookies(callable $match)
+    private function findMatchingCookies(callable $match)
     {
         $cookies = [];
 

      ----------- end diff -----------

   2) src/Encoding/FilteredStream.php
      ---------- begin diff ----------
--- Original
+++ New
@@ -13,12 +13,11 @@
  */
 abstract class FilteredStream implements StreamInterface
 {
-    const BUFFER_SIZE = 8192;
-
     use StreamDecorator {
         rewind as private doRewind;
         seek as private doSeek;
     }
+    const BUFFER_SIZE = 8192;
 
     /**
      * @var callable

      ----------- end diff -----------

```

CHANGELOG.md
peter279k: Thanks. Fixed.

src/CookieJar.php
```diff
@@ -10,7 +10,7 @@
 final class CookieJar implements \Countable, \IteratorAggregate
 {
     /**
-     * @var \SplObjectStorage
+     * @var \SplObjectStorage<object, mixed>
```
peter279k, line 13: Fix static analysis issue via `PHPStan`:

```
 ------ -------------------------------------------------------------- 
  Line   CookieJar.php                                                 
 ------ -------------------------------------------------------------- 
  15     Property Http\Message\CookieJar::$cookies with generic class  
         SplObjectStorage does not specify its types: TObject, TData   
         💡 You can turn this off by setting                            
         checkGenericClassInNonGenericObjectType: false in your        
         phpstan.neon.dist.                                            
 ------ -------------------------------------------------------------- 
```

src/Encoding/FilteredStream.php
```diff
@@ -13,12 +13,11 @@
  */
 abstract class FilteredStream implements StreamInterface
 {
-    const BUFFER_SIZE = 8192;
-
     use StreamDecorator {
         rewind as private doRewind;
         seek as private doSeek;
     }
+    const BUFFER_SIZE = 8192;
```
peter279k, line 20: Fix coding style issue:

```
 2) src/Encoding/FilteredStream.php
      ---------- begin diff ----------
--- Original
+++ New
@@ -13,12 +13,11 @@
  */
 abstract class FilteredStream implements StreamInterface
 {
-    const BUFFER_SIZE = 8192;
-
     use StreamDecorator {
         rewind as private doRewind;
         seek as private doSeek;
     }
+    const BUFFER_SIZE = 8192;
 
     /**
      * @var callable

      ----------- end diff -----------
```

.php_cs.dist
dbu: ah i see. i fixed those method visibilities in the master branch, we can remove the config change.
```suggestion
```

src/Authentication/Wsse.php
```diff
@@ -22,25 +23,39 @@ final class Wsse implements Authentication
...
      * {@inheritdoc}
      */
     public function authenticate(RequestInterface $request)
     {
-        // TODO: generate better nonce?
         $nonce = substr(md5(uniqid(uniqid().'_', true)), 0, 16);
         $created = date('c');
-        $digest = base64_encode(sha1(base64_decode($nonce).$created.$this->password, true));
+        if (false === in_array($this->hashAlgorithm, $this->acceptedHashAlgos)) {
+            throw new InvalidArgumentException(sprintf('Unaccepted hashing algorithm: %s', $this->hashAlgorithm));
```
dbu, line 56: i would not do this. if the user specifies an algorithm that is not supported, there is a clear error message: https://3v4l.org/ju4q7

peter279k: This issue you specify I've resolved. I create the `acceptedHashAlgos` variable and check whether the hashing algorithm is accepted.

.php_cs.dist
peter279k: Thanks. This setting has been removed.

src/Authentication/Wsse.php
dbu: why do you restrict the algorithms to that list? there is [hash_algos](https://www.php.net/manual/en/function.hash-algos.php) that would provide the full list of algorithms that is supported by the installation - but i would not even check the specified algorithm and instead let the `hash` method create the error if something invalid has been specified.

if you want to check with hash_algos, we should do it in the constructor to fail early, at the place where the invalid parameter is passed, rather than some point later.

peter279k: Using [hash_algos](https://www.php.net/manual/en/function.hash-algos.php) function will let this class support many hashing algorithms. Including `md5`, `sha1`, `sha2` and so on.

That's the reason why I use the array to restrict accepted hashing algorithms.

If you have the proper approach, please provide it. Thanks.

peter279k: Or we can leave hashing algorithms once the specific algorithm is valid for `hash` function.

Developers should take their own risks if using the weak hashing algorithms.

And we only do hashing algorithm checking whether it's valid on `hash_algos` during this class constructor.

@dbu, what do you think?

dbu: i was trying to find a proper specification of which algorithms are allowed here, but did not find anything. does WSSE specify which algorithms are allowed?
the only thing i found is https://www.oasis-open.org/committees/download.php/13392/wss-v1.1-spec-pr-UsernameTokenProfile-01.htm and that talks about sha1 as default algorithm or unspecified "other" algorithms. what is the source for the list in acceptedHashAlgos in this PR? if there is some extension to WSSE specifying exactly these algorithms, then the list is of course right to do.

dbu: if technically anything is allowed, i would leave it up to the user of the library. we can do a phpdoc in the constructor argument to list recommended algorithms, but in the end it should be up to the dev. and if you e.g. have to talk with a wsse backend that you have no control over, you might just have to work with a poor algorithm.

peter279k: Thanks. I use the `hash_algos` function to check whether specific hashing algorithm is valis.

dbu (approved):
thanks a lot for the contribution, and thanks for your patience while we sorted this out. this looks ready to merge to me.

@php-http/core any further input or fine if merge and tag version 1.11.0 with this?

dbu:
thanks a lot! i added documentation about this in https://github.com/php-http/documentation/commit/aaeef9cb5ac6e1c2fc1be981c563d5d3dbd907b9 and released this as https://github.com/php-http/message/releases/tag/1.11.0

if you have time to add a configuration option to the symfony bundle as well, that would be neat: https://github.com/php-http/HttplugBundle

PR #139, opened 2021-05-24
support guzzle psr7 version 2

### LINKED ISSUE #138: Bug: Call to undefined function GuzzleHttp\\Psr7\\uri_for()
**PHP version**: 7.4.3

**Description**
After update of GuzzleHttp lib (via composer) , I had this issue:

```
PHP Fatal error:  Uncaught Error: Call to undefined function GuzzleHttp\\Psr7\\uri_for() 
php-http/message/src/UriFactory/GuzzleUriFactory.php

```
**How to reproduce**
Since upgrade (in my case) v2.1.7 dependency lib  GuzzleHttp\\Psr7 to v2.1.7

**Possible Solution**
it seems coming from deprecated use of function : 
php-http\message\src\UriFactory\GuzzleUriFactory.php

original 
`   return Psr7\uri_for($uri);`
#
correction : 
`return Psr7\Utils::uriFor($uri);`

I only fast edit this fix. I didn't check all deprecated functions used.

dbu:
| Q               | A
| --------------- | ---
| Bug fix?         | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #138
| License         | MIT

#### What's in this PR?

Support guzzle psr7 version 2 that moved functions to static methods of the Utils class. Our factory is deprecated but still used when no other factory is available, and as we don't restrict the guzzle psr version, we end up failing.

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -31,6 +31,6 @@ jobs:
         uses: actions/checkout@v2
 
       - name: PHP-CS-Fixer
-        uses: docker://oskarstark/php-cs-fixer-ga
+        uses: docker://oskarstark/php-cs-fixer-ga:2.19.0
         with:
           args: --dry-run --diff-format udiff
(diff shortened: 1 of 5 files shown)


dbu:
@Nyholm ok if i delete the scrutinizer webhook? it fails with something that looks like an error inside scrutinizer. and with php-cs and phpstan i think we are covered well enough.

Nyholm:
Yes, lets!

dbu:
and done.

i tagged 1.11.1 with the forward-compatibility fix.

PR #146, opened 2022-02-07
Add Formatter::formatResponseForRequest()

### LINKED ISSUE #145: Expose request to resonse formatter
Hi,

Would you consider passing the request to `formatResponse()` in `Formatter`? Introducing "format response for request" effectively.

We could use it to redact sensitve responses for certain URLs.

We could use it to hide response bodies for certain URLs that we don't care about.

The alternative route is to create a formatter per usecase, thus a plugin client per usecase ... which is really overhead :)

WDYT?

ro0NL:
| Q               | A
| --------------- | ---
| Bug fix?         | no|
| New feature?    | yes
| BC breaks?      | no|
| Deprecations?   | yes |
| Related tickets | fixes #145
| Documentation   | -
| License         | MIT

During 1.x lifecycle:

```
deprecation.INFO: User Deprecated: Class "Http\HttplugBundle\Collector\Formatter" should implement method "Http\Message\Formatter::formatResponseForRequest(ResponseInterface $response, RequestInterface $request): string": Formats a response in context of its request.
```

Next major release should add the method for real.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.13.0] - tomorrow
+
+- Added `Formatter::formatResponseForRequest()`
+
 ## [1.12.0] - 2021-08-29
 
 - Added support for adjusting binary detection regex in FullHttpMessageFormatter
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [1.13.0] - tomorrow
 
 - Added `Formatter::formatResponseForRequest()`
+- Deprecated `Formatter::formatResponse()`
 
 ## [1.12.0] - 2021-08-29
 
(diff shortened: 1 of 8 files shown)


dbu (approved):
looks good to me.

@Nyholm wdyt?

there is some code rot going on with php 8 and phpstan, but thats not related to the PR.

Nyholm:
Im happy with this. 

But this will only have a "real" effect in v2. Currently v2 is not planned. 

Introducing a new interface would make more sense and would make this change "real" now. I see that it was discussed in #145

ro0NL:
i will update at least logger plugin as such, using method_exsists trick till v2.

im still not convinced we need moar interfaces :')

dbu:
> But this will only have a "real" effect in v2. Currently v2 is not planned.

aparently this is the symfony way of "adding" methods to interfaces without breaking BC. the consumer would have to use method_exists as i understood

dbu:
i agree with @ro0NL's reasoning. ok for you @Nyholm or do you have a strong feeling that that would be wrong?

ro0NL:
about deprecating formatResponse(): https://github.com/php-http/message/issues/145#issuecomment-1030182461

i think it makes sense :+1: and ideally there's 1 responsible for method for response formatting. Let me know if should be part of this PR.

dbu:
if we do the new method on the same interface, i would indeed deprecate the old formatResponse method. and best to this in the same MR.

dbu:
thanks! i will see if we can get the build to green again, that would be good to see before tagging the release.

dbu:
and released https://github.com/php-http/message/releases/tag/1.13.0

PR #147, opened 2022-02-10
adjust baseline to reformatted error messages and fix some issues

dbu:


dbu:
@GrahamCampbell in 075b945c7018f65004cdccf1dd8755e3d79af971 you adjusted the FilteredStreamStubSpec for a different behaviour with PHP 8. did that behaviour later get reverted? the error we get seems that in PHP 8 a RuntimeException is thrown and not a phpspec ErrorException. would be great if you could have a look at it, would love to get our tests on this repo to be green again :-)

PR #150, opened 2023-04-11
Allow "psr/http-message" v2

nicolas-grekas:
| Q               | A
| --------------- | ---
| Bug fix?         | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "php": "^7.1 || ^8.0",
         "clue/stream-filter": "^1.5",
         "php-http/message-factory": "^1.0.2",
-        "psr/http-message": "^1.0"
+        "psr/http-message": "^1.0 || ^2.0"
     },
     "provide": {
         "php-http/message-factory-implementation": "1.0"


dbu:
fixing all on the internet will take a little while i fear. we don't actually test with PSR-7 2.0 yet. when i try to force 2.0, i get:
```
  Problem 3
    - guzzlehttp/psr7 is locked to version 1.9.0 and an update of this package was not requested.
    - guzzlehttp/psr7 1.9.0 requires psr/http-message ~1.0 -> found psr/http-message[1.0, 1.0.1, 1.1] but it conflicts with your root composer.json require (^2.0).
  Problem 4
    - slim/slim is locked to version 3.12.4 and an update of this package was not requested.
    - slim/slim 3.12.4 requires psr/http-message ^1.0 -> found psr/http-message[1.0, 1.0.1, 1.1] but it conflicts with your root composer.json require (^2.0).
  Problem 5
    - laminas/laminas-diactoros is locked to version 2.25.1 and an update of this package was not requested.
    - laminas/laminas-diactoros 2.25.1 requires psr/http-message ^1.1 -> found psr/http-message[1.1] but it conflicts with your root composer.json require (^2.0).
```

however, we don't implement any of these interfaces in this repository, therefore i think it is safe enough to merge this.

Issue #158, opened 2024-01-23
CI: Add testrun for "psr/http-message" v2 with slim


Chris53897:
in composer.json "psr/http-message" v1 and v2 are allowed.

But if i see this correct, v2 is never tested in the testruns.
Example:
https://github.com/php-http/message/actions/runs/7623563875/job/20763848547

Because of Version 3 of "slim/slim"
https://github.com/php-http/message/blob/1.x/composer.json#L31C10-L31C19

https://packagist.org/packages/slim/slim#3.12.5

Only the dev-version supports v1 and v2
https://packagist.org/packages/slim/slim#4.x-dev

dbu:
hm, tricky :-/ 

maybe we should remove these optional integrations from require-dev and only require them for specific test runs. we could tag the tests about the integrations and only run tests with that tag in a build where we requite them. that would make the rest of the tests run with the latest and greatest for proper testing.

Chris53897:
I just noticed that there is a testrun for php 8.2 where "slim/slim" is removed in ci.
https://github.com/Chris53897/message/actions/runs/7624004757/job/20765252556

But maybe with the latest dev-version of "slim/slim" this could be run on every php version.

PR #159, opened 2024-01-23
chore: add testrun for "psr/http-message" v2

Chris53897:
see https://github.com/php-http/message/issues/158

Just for testing the CI

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,31 +31,6 @@ jobs:
       - name: Execute tests
         run: composer test
 
-  psr-7_2:
-    name: PHP PSR-7 2.0
-    runs-on: ubuntu-latest
-
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
-
-      - name: Setup PHP
-        uses: shivammathur/setup-php@v2
-        with:
-          php-version: 8.2
-          tools: composer
-          coverage: none
-
-      - name: Install dependencies
-        run: |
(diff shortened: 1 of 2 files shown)


Chris53897:
I do not know how to fix the changes between slim 3 and 4.
Did not find an easy migration path.

dbu:
i am a bit afraid of testing with a dev version.

could we just remove slim from require-dev and add it with composer commands in the github ci for some builds but not all? and have the tests that depend on slim skip if slim is not available?

Chris53897:
Yes of course. Maybe they will release a new version?
https://github.com/slimphp/Slim/issues/3311

But i would suggest that the breaking-changes should be fixed first.
At best from someone with more knowledge at this topic.

dbu:
slim has tagged a release. can you please rebase this branch onto 1.x and push again? then we can see what problems remain.

Chris53897:
@dbu Thanks
Now i can work on the support for slim 4.

.github/workflows/tests.yml
```diff
@@ -31,31 +31,6 @@ jobs:
...
-        with:
-          php-version: 8.2
-          tools: composer
-          coverage: none
-
-      - name: Install dependencies
-        run: |
-          rm src/MessageFactory/SlimMessageFactory.php src/StreamFactory/SlimStreamFactory.php src/UriFactory/SlimUriFactory.php spec/MessageFactory/SlimMessageFactorySpec.php spec/StreamFactory/SlimStreamFactorySpec.php spec/UriFactory/SlimUriFactorySpec.php
-          composer remove --dev "slim/slim" --no-interaction --no-update
-          composer require "psr/http-message:^2.0" --no-interaction --no-update
```
dbu, line 53: should we not keep this test around?

dbu (commented):
@Chris53897 do you have updates on this one? anything blocking, or did you just not find time to continue?

PR #166, opened 2024-10-14
feat: remove ergebnis/composer-normalize

Chris53897:
I want to test CI if is needed "ergebnis/composer-normalize" while it runs via docker gh actions

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,6 @@
     },
     "require-dev": {
         "ext-zlib": "*",
-        "ergebnis/composer-normalize": "^2.6",
         "guzzlehttp/psr7": "^1.0 || ^2.0",
         "php-http/message-factory": "^1.0.2",
         "phpspec/phpspec": "^5.1 || ^6.3 || ^7.1",


dbu:
it seems to work. but we have some dev dependencies in composer.json require-dev so that people can run them locally, same with phpspec. 

is there an issue with having normalize in the require-dev section?

Chris53897:
Only deprecations in the testruns for PHP 8.4.
Example: https://github.com/php-http/message/actions/runs/11143029675/job/30967217820

I have created a PR.
https://github.com/localheinz/diff/pull/72

But i thought maybe it is not needed anymore, as it runs via docker gh actions.

dbu:
@Chris53897 the fix to localheinz/diff has been merged and tagged. did that solve the deprecations, can we close this MR?

Chris53897:
Thanks for the reminder.
It looks good. https://github.com/php-http/message/actions/runs/13344353123/job/37272964327?pr=167
I will clsoe this PR.

PR #167, opened 2025-02-15
docs: remove $ in bash command

Chris53897:
test for https://github.com/php-http/message/pull/166

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 Via Composer
 
 ``` bash
-$ composer require php-http/message
+composer require php-http/message
 ```
 
 
@@ -37,7 +37,7 @@ Please see the [official documentation](http://docs.php-http.org/en/latest/messa
 ## Testing
 
 ``` bash
-$ composer test
+composer test
 ```
 
 


Chris53897:
With the removed `$` the code can be run within the IDE or and ist better for copy-paste

dbu:
looks good right?

the phpstan errors are probably because of improved phpstan, and we seem to miss the configuration that it should not treat phpdoc as guaranteed.

Chris53897:
Yes. phpstan is not related to this PR.Could be fixed in a seperate PR.Am 16.02.2025 um 17:09 schrieb David Buchmann ***@***.***>:﻿
  looks good right?
the phpstan errors are probably because of improved phpstan, and we seem to miss the configuration that it should not treat phpdoc as guaranteed.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>



dbu left a comment (php-http/message#167)
looks good right?
the phpstan errors are probably because of improved phpstan, and we seem to miss the configuration that it should not treat phpdoc as guaranteed.

dbu:
yeah phpstan is not related. i checked the normalize output and saw no deprecation warnings.

